PR #5, opened 2017-06-21
Docker Setup

jameeters:


Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    python3-pip \
+    nginx-light \
+    curl \
+    apt-transport-https \
+    git
+
+COPY . /app
+WORKDIR /app
+
+RUN cp DOIRootCA2.cer /etc/ssl/certs/DOIRootCA2.pem \
```
ayan-usgs: I'm not sure this will have the desired effect of copying the root DOI certificate from the host to the Docker container. This command will copy a `DOIRootCA2.cer` that's already in the container to `/etc/ssl/certs/DOIRootCA2.pem`. Using `COPY` should give the desired effect.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+&& cat DOIRootCA2.cer >> /etc/ssl/certs/ca-certificates.crt
+
+RUN rm -f /etc/apt/sources.list.d/chris-lea-node_js-jessie.list \
+&& curl --cacert DOIRootCA2.cer https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - \
+&& echo 'deb https://deb.nodesource.com/node_6.x jessie main' > /etc/apt/sources.list.d/nodesource.list \
+&& apt-get update \
+&& apt-get install -y nodejs
+
+RUN pip3 install -r requirements.txt
```
ayan-usgs: I expect this command to run into problems when downloading packages from PyPI because `pip3` won't know about the DOI root certificate. Information regarding pip configuration can be found at https://pip.pypa.io/en/stable/user_guide/#configuration.

Dockerfile
jameeters: I copy the certificate into the container with `COPY . /app`. I was under the impression that I needed to store the certificate with the system certificates, but everything seems to work fine with these lines commented out.

Dockerfile
jameeters: I thought that this was working because I was storing the DOI cert with the system certs, but it still works with those lines commented out and without pointing it at the DOI certificate. I think I'll add a `--cert` flag just to be safe.

Dockerfile
ayan-usgs: Ah. That makes sense. I would recommend retaining that line then.

PR #6, opened 2017-06-22
Hydrograph with NVD3 and Flask data processing 

ede0m:
Configurable parameters can be set in instance/config.py


floodviz/views.py
```diff
@@ -1,7 +1,79 @@
...
+				date = dt.split('T')[0]
+				t = dt.split('T')[1].split('.')[0]
+			 	# reformat datetime for python datetime
+				dt = datetime.strptime(date + ' ' + t, '%Y-%m-%d %H:%M:%S')
+				# Conver to miliseconds for use with d3 x axis format
+				dt_ms = time.mktime(dt.timetuple()) * 1000
+				# create dummy value for nvd3 issue at https://github.com/novus/nvd3/issues/695 #
+				if idx2 is 0:
+					newd[idx]['values'].append({'date':start_date, "time": 0, "time_mili": dt_ms, 'value': 0, 'max': max_val})
+				# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
```
ede0m: There is an issue within NVD3 and cumulativeLineCharts that normalizes the Y axis data values incorrectly, thus misrepresenting them on the Y axis. In the link in the comment, NatasaPeic presents a workaround in which you create a dummy data value of 0 for only the first data point. After putting that in, the graph renders correctly.

mbucknell: Maybe in the first comment explain that it applies to the if statement below. I just wasn't sure what the line of "###....#" meant.

floodviz/views.py
```diff
@@ -1,7 +1,80 @@
...
+    sites_value_maxs = {}
+    sites = app.config['SITE_IDS']
+    start_date = app.config['START_DT']
+    end_date = app.config['END_DT']
+    n_show_series = app.config['N_SERIES']
+    hydro_meta = app.config['HYDRO_META']
+
+    # Set up to retrieve all site ids from url? #
+    for idx, site in enumerate(sites):
+        url = 'https://nwis.waterservices.usgs.gov/nwis/iv/?site=' + site + '&startDT=' + \
```
mbucknell: You should put this service endpoint in the main config.py. Probably only https://nwis.waterservices.usgs.gov/nwis as the rest could vary if we decide to pull down different kinds of data.

floodviz/views.py
```diff
@@ -1,7 +1,79 @@
...
+	n_show_series = app.config['N_SERIES']
+	hydro_meta = app.config['HYDRO_META']
+
+	## Set up to retrieve all site ids from url? ##
+	for idx, site in enumerate(sites):
+		url = 'https://nwis.waterservices.usgs.gov/nwis/iv/?site=' + site + '&startDT=' + start_date + '&endDT=' + end_date + '&parameterCD=00060&format=json'
+
+		try:
+			r = requests.get(url)
+		except:
```
ede0m: Was just using that intermediately for overall fault tolerance. But later added the response code if statement and did not remove the try. I must see them too much and glanced right over.

floodviz/views.py
```diff
@@ -1,7 +1,79 @@
...
+			print('Unable to retrieve URL: ' + url)
+			continue
+		
+
+		if r.status_code is 200:
+			j = r.json()
+			siteName = j['value']['timeSeries'][0]['sourceInfo']['siteName']
```
ede0m: I like this idea. I will refactor to make it happen.

PR #7, opened 2017-06-22
Map Creation Demo

skaymen:
Currently generates an SVG map-- which must be viewed on its own. Not currently integrated with the HTML.


floodviz/static/data/cities.json
```diff
@@ -0,0 +1,32 @@
+
```
mbucknell: This should not be committed. Is this the geojson that Lindsay gave us? This should be in instance/config.py. Could be represented with a single string variable that contains the json string. 

All config variables that we define should have defaults (the empty string for this one would be good) and a comment describing what you expect.

floodviz/views.py
```diff
@@ -1,7 +1,112 @@
 from flask import request, render_template, abort
+import csv
```
mbucknell: See https://www.python.org/dev/peps/pep-0008/#imports for order of imports. Within a group of imports, I like to put them in alphabetical order.

floodviz/views.py
```diff
@@ -1,7 +1,112 @@
...
 
+
 @app.route('/home/')
 def home():
-    return render_template('index.html')
\ No newline at end of file
+    return render_template('index.html')
+
+
+# This program requires svgis to be installed (which in turn requires Fiona and GDAL)
```
mbucknell: I'm sorry that I led you down the wrong path here. I was hoping that all of our preprocessing steps would be within the view itself. The package svgis does have an API so it should be possible to create the SVG when the view is rendered.  I would like you to see if that is possible. Also did you look at using D3 to draw the map client side? Just wondering if you explored that.

floodviz/static/data/cities.json
skaymen: These are the cities given by Jordan in the "issues" tab on GitHub. I will modify the code so that they are contained in a variable instead.

floodviz/views.py
skaymen: I will correct this.

floodviz/views.py
skaymen: I will explore using the svgis API. 

I briefly looked at D3 but went with svgis as it was simpler to use and seemed to suit our purpose. If svgis continues to cause problems I am willing to switch to using D3.

mbucknell: svgis may be simpler to use but there are also additional libraries that have to be installed. I think it's something we will need to talk about with the data science folks. It will be interesting to see the SVG that is produced by this project. We will talk more about this at the demo tomorrow.

floodviz/views.py
```diff
@@ -1,7 +1,112 @@
...
+#     for site in siteReader:
+#         if site[0].isalnum():
+#             siteList.append(site[0])
+
+
+# generate the string of site ids for the url
+idInputString = ",".join(siteList)
+# print(idInputString)
+
+url = "https://waterservices.usgs.gov/nwis/site/?format=rdb&sites=" + idInputString + "&siteStatus=all"
```
mbucknell: You should use the config variable app.config['NWIS_SITE_SERVICE_ENDPOINT here.

floodviz/views.py
```diff
@@ -1,7 +1,112 @@
...
+
+# get data from url
+req = requests.get(url)
+
+# Discard preamble from response
```
mbucknell: I suspect this will not work in some cases. If any of the data in the columns contains a '#' sign, you will have removed too much. It may be better to iterate over the lines until you get to the first one that does not start with #.

I would also put this into a separate function within a utils module. Seems like something we might want to reuse and it would be good to write unit tests for it.

floodviz/static/data/cities.json
skaymen: Fixed. I used a dict instead of a string to preserve the double quotes.

floodviz/views.py
skaymen: Fixed.

floodviz/static/data/cities.json
mbucknell: Just FYI, you can use single quotes around a python string and therefore preserve the double quotes.

README.md
```diff
@@ -1 +1,8 @@
-# active-flood-viz
\ No newline at end of file
+# active-flood-viz
+
+SVGIS requires Fiona, which in turn requires GDAL, which in turn requires the Visual C++ Build Tools -- which must be
+downloaded from http://landinghub.visualstudio.com/visual-cpp-build-tools
+
+Also, on my machine, it takes about 60 seconds for the SVG to actually be generated.
```
mbucknell: These comments (lines 6 and 8) don't belong in a README. README's should contain information about how to set up the project, so lines 3 and 4 are appropriate. Comments about how to run the application may also be included. And descriptions of what the repo does. Configuration parameter description may also be included.

floodviz/views.py
```diff
@@ -1,7 +1,112 @@
...
+for line in data[1:]:
+    line = line.split('\t')
+    line_dict = dict(zip(fields, line))
+    data_nice.append(line_dict)
+
+# keep only station name, id, huc, lat and lon
+arr = []
+
+# store this data in arr
+for datum in data_nice:
```
mbucknell: Is there a reason for changing the data into an array of arrays? Would it be clearer to convert the array of dictionaries into a geojson rather than converting to an array of arrays?

floodviz/views.py
```diff
@@ -1,7 +1,112 @@
...
+    if 'station_nm' in datum.keys():
```
mbucknell: What is the purpose of this if? Why only check for 'station_nm'? is the check needed at all? You could use datum.get('station_nm', '') which would handle the situation of station_nm missing.

floodviz/views.py
```diff
@@ -1,7 +1,112 @@
...
+        temp = [datum['station_nm'], datum['site_no'], datum['huc_cd'], datum['dec_lat_va'], datum['dec_long_va']]
+        arr.append(temp)
+
+# write data to sites.json in geojson format
+with open("floodviz/static/data/gages.json", "w") as f:
```
mbucknell: I'd like to see this as a separate function.

floodviz/views.py
skaymen: Have switched code over to svgis API.

floodviz/views.py
skaymen: Fixed.

floodviz/views.py
mbucknell: Cool. Looking forward to seeing the code.

README.md
skaymen: Fixed.

floodviz/views.py
```diff
@@ -105,8 +113,13 @@ def home():
...
-          "-i \"name\" -c floodviz/static/css/site_style.css -b " + \
-          str(bound1) + " " + str(bound2) + " " + str(bound3) + " " + str(bound4)
-os.system(command)
+# command = "svgis draw floodviz/static/data/gages.json floodviz/static/data/cities.json -o " \
+#           "floodviz/static/data/sitemap.svg -f 500 -j epsg:2794 " \
+#           "-i \"name\" -c floodviz/static/css/site_style.css -b " + \
+#           str(bound1) + " " + str(bound2) + " " + str(bound3) + " " + str(bound4)
+# os.system(command)
+
+x = svgis.map("floodviz/static/data/gages.json", scale=300, crs="epsg:2794", bounds=bounds)
```
mbucknell: Actually that was pretty easy! But it still needs to happen when the view is rendered. I realize that it will slow down the rendering of the page, but that's why we will be freezing the app.

floodviz/views.py
skaymen: Fixed.

floodviz/views.py
skaymen: Fixed.

floodviz/views.py
skaymen: Fixed.

floodviz/views.py
skaymen: Fixed.

floodviz/views.py
skaymen: Fixed.

PR #8, opened 2017-06-22
Frozen flask setup

jameeters:


run.py
```diff
@@ -11,6 +13,6 @@
         host = host_val
     else:
         host = '127.0.0.1'
-    application.run(host=host, port=5050, threaded=True)
+    freezer.run(host=host, port=5050, threaded=True)
```
mbucknell: We want to be able to preserve the ability to run either the application dynamically or to freeze it. This may involve using something like Flask-Script to allow us to create  commands. See https://github.com/USGS-CIDA/WQP_UI/blob/master/manage.py for an example of how we are doing this for other flask tools. I

README.md
```diff
@@ -10,6 +10,6 @@ Currently, our dockerfile creates an image for a container that runs the basic f
 1. Run `docker build -t flood .`. This will create the docker image and name it `flood`.
 
 **To run a container based on the image:**
-1. `docker run -p 5055:5050 --name floodviz flood`. This will create a docker container from the `flood` image and name it `floodviz`. 
+1. `docker run -p 5055:5050 -t --name floodviz flood`. This will create a docker container from the `flood` image and name it `floodviz`.
```
mbucknell: Should the "-t" be "-f" in order to freeze the application?

jameeters: The `-t` tells docker to a terminal output for the container, without this it won't print "frozen" so it's hard to tell if it actually did freeze. When you start the container it runs the command at the end of the dockerfile, which is `python3 run.py --host 0.0.0.0 --freeze`.

mbucknell: Gotcha! Is this ready to merge?

PR #9, opened 2017-06-22
Hydrograph with NVD3 and Flask Data Processing -- 2

ede0m:


floodviz/hydrograph_utils.py
```diff
@@ -0,0 +1,42 @@
+from datetime import datetime
+import operator
+import time
+
+
+def parse_hydrodata(jdata, newd, idx, sites_value_maxes):
```
mbucknell: Need to add comments describing the function, the inputs, and what it returns.Generally we use sphinx standards. You can go here for an example: https://pythonhosted.org/an_example_pypi_project/sphinx.html#function-definitions

It is preferable to write functions that don't have side effects - in other words they don't change their inputs or global data. For this function, I was expecting it to return the parsed data. Looking at it more closely it looks like it's parsing and returning a max val. Does it make sense to split these into two functions? If not consider returning a dictionary containing a key for the parsed data and a key for the max value.

floodviz/hydrograph_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    siteName = jdata['value']['timeSeries'][0]['sourceInfo']['siteName']
```
mbucknell: Rather than passing in the whole jdata dictionary, consider passing in just jdata['value']'[timeSeries'][0]

floodviz/hydrograph_utils.py
```diff
@@ -0,0 +1,42 @@
...
+        t = dt.split('T')[1].split('.')[0]
+        # reformat datetime for python datetime #   
+        dt = datetime.strptime(date + ' ' + t, '%Y-%m-%d %H:%M:%S')
+        # Convert to milliseconds for use with d3 x axis format
+        dt_ms = time.mktime(dt.timetuple()) * 1000
+        # (for below if statment) create dummy value for nvd3 issue at https://github.com/novus/nvd3/issues/695 #
+        if idx2 is 0: # First datapoint of this site
+            newd[idx]['values'].append({'date': date, "time": 0, "time_mili": dt_ms, 'value': 0, 'max': max_val})
+   
+        newd[idx]['values'].append({'date': date, "time": t, "time_mili": dt_ms, 'value': value, 'max': max_val})
```
mbucknell: Why is the current max_val appended to the returned parsed data?

floodviz/hydrograph_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+    sites_value_maxes[key] = max_val
+
+    
+def filter_hydrodata(newd, sites_value_maxes, n_show_series):
```
mbucknell: Add comments and make the function side effect free with a returned value.

floodviz/static/js/hydrograph.js
```diff
@@ -0,0 +1,55 @@
+document.addEventListener("DOMContentLoaded", function(event) {
```
mbucknell: If we start doing any serious DOM selection/manipulation or adding event handlers, I think we should start using jquery, but I'm fine with leaving this as is for now.

floodviz/static/js/hydrograph.js
```diff
@@ -0,0 +1,55 @@
...
+					.x( getX )  // this value is stored in milliseconds since epoch (converted in data_format.py with datetime)
+					.y( getY ) 
+					.color(d3.scale.category10().range())
+					.useInteractiveGuideline(true)
+					.yDomain([min, max])
+					.margin({left: 120, top: 60})
+					.showControls(false);
+
+			chart.xAxis
+					.axisLabel(" Date (M-D-Y)")
```
mbucknell: Let's follow the pattern that you established with the first chained call in line 22. So line 31 should be chart.xAxis.axLabel....

floodviz/templates/index.html
```diff
@@ -1,14 +1,35 @@
...
 <html lang="en">
+
 <head>
 	<meta charset="UTF-8">
-	<link type="text/css" href="{{ url_for('static', filename='bower_components/bootstrap/dist/css/bootstrap.css') }}" />
-	<title></title>
+	<link rel="stylesheet" type="text/css" href="{{ url_for('static', filename='bower_components/bootstrap/dist/css/bootstrap.css') }}" />
+	<link rel="stylesheet" type="text/css" href="{{ url_for('static', filename='bower_components/nvd3/build/nv.d3.css') }}" />
+	<link rel="stylesheet" type="text/css" href="{{ url_for('static', filename='css/style.css') }}"/>
+	<script type="text/javascript" src="{{ url_for('static', filename='bower_components/jquery/dist/jquery.min.js') }}"></script>
```
mbucknell: Why did you move the script tags for jquery and bootstrap?

floodviz/templates/index.html
```diff
@@ -1,14 +1,35 @@
...
+	<script type="text/javascript" src="{{ url_for('static', filename='bower_components/bootstrap/dist/js/bootstrap.js') }}"></script>
+
+	<title>Flood Viz</title>
+
 </head>
+
 <body>
-	<div class="main-content container-fluid">This is where the visualization will go</div>
+	
+	<script type="text/javascript">
```
mbucknell: This should go below all of the html content but before any of the other script tags.

floodviz/views.py
```diff
@@ -1,7 +1,41 @@
...
+    sites = app.config['SITE_IDS']
+    start_date = app.config['START_DT']
+    end_date = app.config['END_DT']
+    n_show_series = app.config['N_SERIES']
+    hydro_meta = app.config['HYDRO_META']
+    url_top = app.config['NWIS_SITE_SERVICE_ENDPOINT']
+    sites_value_maxes = {}
+
+    # Set up to retrieve all site ids from url? #
+    for idx, site in enumerate(sites):
```
mbucknell: Can the data for all of the sites be retrieved with a single call? I think you can use the sites=x1,x2,x3 parameter. That's preferable to making multiple calls. Parsing may be more difficult but that's preferable to making multiple requests.

floodviz/views.py
```diff
@@ -1,7 +1,41 @@
...
+        url =  url_top +'iv/?site=' + site + '&startDT=' + \
+              start_date + '&endDT=' + end_date + '&parameterCD=00060&format=json'
+
+        r = requests.get(url)
+
+        if r.status_code is 200:
```
mbucknell: We should think about what to do in case the status code is not 200. One easy thing to do would be to write the file out with an empty object. This may be what happens below. I'm not sure but it is something to consider and perhaps at least do a test where you get an error and see what happens in the application.

floodviz/hydrograph_utils.py
ede0m: Was using for a .forceY call I was experimenting with in hydrograph.js but later removed.

floodviz/views.py
ede0m: Done. Good call

floodviz/hydrograph_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def parse_hydrodata(jdata):
+    
+    """ 
+    Parses json Hydrodata from NWIS webservice
+    and formats for NVD3 charting library.
+
+    ARGS: 
+        jdata - json containing time series data for all sites listed in config.py SITE_IDs
```
mbucknell: Should be specific about what types you are expecting. For instance this is a list of dict

floodviz/hydrograph_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    
+    RETURNS:
+        all series data correctly formated for NVD3 charting library.
```
mbucknell: Be specific about what type you are returning

floodviz/templates/index.html
mbucknell: This needs to be addressed before merging.

floodviz/views.py
```diff
@@ -1,7 +1,44 @@
...
+    sites = app.config['SITE_IDS']
+    start_date = app.config['START_DT']
+    end_date = app.config['END_DT']
+    n_show_series = app.config['N_SERIES']
+    hydro_meta = app.config['HYDRO_META']
+    url_top = app.config['NWIS_SITE_SERVICE_ENDPOINT']
+
+    # Set up to retrieve all site ids from url #
+    site_string = ''
+    for idx, site in enumerate(sites):
```
mbucknell: Take a look at the string join method. It will reduce this for loop to a single statement.

floodviz/templates/index.html
ede0m: Moved back down to the bottom of the body.

floodviz/templates/index.html
```diff
@@ -29,6 +24,9 @@
 
 	<script type="text/javascript" src="{{ url_for('static', filename='bower_components/jquery/dist/jquery.min.js') }}"></script>
 	<script type="text/javascript" src="{{ url_for('static', filename='bower_components/bootstrap/dist/js/bootstrap.js') }}"></script>
+	<script type="text/javascript" src="{{ url_for('static', filename='js/hydrograph.js')}}"></script>
```
mbucknell: Sorry, this should go after the d3 and nvd3 because the code depends on it. Generally it's best to maintain that kinds of ordering.

ede0m: Done

mbucknell:
Going to merge. Still need to add unit tests for utils.

PR #11, opened 2017-06-23
Added maputils.py so the functions can be used in the future.

skaymen:
Contains maputils.py.


floodviz/maputils.py
```diff
@@ -0,0 +1,58 @@
...
+
+def site_dict(site_list, url_prefix):
+    """Puts site data into a dictionary
+
+    Args:
+        site_list: A list of site ids to be queried on NWIS
+        url_prefix: The beginning of the NWIS url
+
+    Returns:
+        A dict containing various site information in a usable format.
```
mbucknell: You are actually returning an array of dict.

floodviz/maputils.py
```diff
@@ -0,0 +1,58 @@
+import requests
+
+def site_dict(site_list, url_prefix):
+    """Puts site data into a dictionary
+
+    Args:
+        site_list: A list of site ids to be queried on NWIS
+        url_prefix: The beginning of the NWIS url
```
mbucknell: You should also describe the type of url_prefix -- so something like "A string containing the NWIS site url".

floodviz/maputils.py
```diff
@@ -0,0 +1,58 @@
...
+        dnice.append(line_dict)
+
+    return dnice
+
+
+def write_geojson(filename, data):
+    """Writes site data to a .json file so it can be mapped
+
+       Args:
+           filename: the file to be written to
```
mbucknell: Please describe the parameter types. Particularly important for data which is an array of dict.

floodviz/maputils.py
```diff
@@ -0,0 +1,58 @@
...
+
+    with open(filename, "w") as f:
+        f.write("{ \"type\": \"FeatureCollection\", \"features\": [ \n")
+
+        for i, datum in enumerate(data):
+            f.write("{ \"type\": \"Feature\",\n \"geometry\": {\n \"type\": \"Point\",\n \"coordinates\" : "
+                    "[" + datum.get('dec_long_va') + ", " + datum.get('dec_lat_va') + "]\n },\n")
+            f.write(" \"properties\": {\n \"name\": \"" + datum.get('station_nm') + "\",\n \"id\": \""
+                    + datum.get('site_no') + "\",\n \"huc\": \"" +
+                    datum['huc_cd'] + "\" \n } \n }")
```
mbucknell: Should use get for this one too.

floodviz/maputils.py
skaymen: Fixed.

floodviz/maputils.py
skaymen: Fixed.

floodviz/maputils.py
skaymen: Fixed.

floodviz/maputils.py
skaymen: Fixed.

PR #16, opened 2017-06-27
Unit Tests and Refactoring

skaymen:
Refactored code in map_utils based on feedback as well as adding some basic error handling. Wrote a number of unit tests.


floodviz/map_utils.py
```diff
@@ -2,15 +2,33 @@
...
+
+    for site in site_list:
+        if not site.isdigit():
+            print("Error: list contains a non-numeric element")
+            return
+        if len(site) != 8:
+            print("Error: all site ids must be 8 digits")
+            return
+
+    if not url_prefix == 'https://waterservices.usgs.gov/nwis/site':
```
mbucknell: This is unnecessary. The point of passing in the url_prefix is so that the endpoint can be changed.

floodviz/map_utils.py
```diff
@@ -20,8 +38,19 @@ def site_dict(site_list, url_prefix):
     # get data from url
     req = requests.get(url)
```
mbucknell: Still need to handle non 200 status code responses. In this case, I would return None and print out a message. Your function comment should reflect that if the service call fails that the function will return None.

floodviz/map_utils.py
```diff
@@ -2,15 +2,33 @@
...
         site_list: A list of site ids to be queried on NWIS
-        url_prefix: The beginning of the NWIS url
-
+        url_prefix: A string containing the beginning of the NWIS site url
     Returns:
-        A dict containing various site information in a usable format.
+        An array of dicts containing various site information in a usable format.
     """
 
+    if not site_list:
```
mbucknell: So your function comments say that it returns an array of dictionaries. If an empty list of sites is passed in, then return an empty array. I don't think this is actually an error condition. You could print out an info statement but it's not really an error.

floodviz/map_utils.py
```diff
@@ -2,15 +2,33 @@
...
+        print("Error: site_list is empty")
+        return
+
+    if not url_prefix:
+        print("Error: url_prefix is empty")
+        return
+
+    for site in site_list:
+        if not site.isdigit():
```
mbucknell: The service call itself will fail if the sites are invalid. So rather than do this checking here make sure you handle non 200 status codes in req.status_code. This applies to both of these if statements which should be eliminated.

floodviz/map_utils.py
```diff
@@ -20,8 +38,19 @@ def site_dict(site_list, url_prefix):
     # get data from url
     req = requests.get(url)
 
+    if not req.text.startswith('#'):
```
mbucknell: This is not necessary. It's possible that an rdb file would contain no comment lines.

floodviz/map_utils.py
```diff
@@ -20,8 +38,19 @@ def site_dict(site_list, url_prefix):
...
+        return
+
     # put data into list
-    data = req.text.splitlines()[29:]
+    data = []
+    for line in req.text.splitlines():
+        if not line.startswith('#'):
+            data.append(line)
+
+    if not data:
```
mbucknell: It's possible that your call won't return any data. It's not an error so this should not print out an error and instead return an empty array.

floodviz/map_utils.py
```diff
@@ -20,8 +38,19 @@ def site_dict(site_list, url_prefix):
...
 
+    if not req.text.startswith('#'):
+        print("Error: request returns invalid data")
+        return
+
     # put data into list
-    data = req.text.splitlines()[29:]
+    data = []
+    for line in req.text.splitlines():
+        if not line.startswith('#'):
```
mbucknell: This will eliminate any lines that start with #. The rdb standard says that the first line without a # starts the file. It's possible that a data column might start with a #. This would erronously omit it. You should write this so once the first non comment line is found, the remaining lines are data.

floodviz/map_utils.py
```diff
@@ -36,11 +65,29 @@ def site_dict(site_list, url_prefix):
...
 def write_geojson(filename, data):
     """Writes site data to a .json file so it can be mapped
-
        Args:
-           filename: the file to be written to
-           data: the data to be written to the file
+           filename: A string naming the file to be written to
+           data: A list of dicts with data to be written to the file
     """
+    if not type(filename) == str:
```
mbucknell: These kind of checks are overkill. The user of this function knows to pass in a string and it does not need to end with .json.

floodviz/map_utils.py
```diff
@@ -36,11 +65,29 @@ def site_dict(site_list, url_prefix):
...
+        print("Error: filename must be string")
+        return;
+
+    if not filename.endswith(".json"):
+        print("Error: outfile must be .json")
+        return
+
+    if not data:
```
mbucknell: Similarly, the user of the function know data is a list. If the list is empty then write an empty array to the json file. FYI the first if statement would be true if data is None so the second if statement would never be executed. All three of these if statements are not needed.

floodviz/map_utils_tests.py
```diff
@@ -0,0 +1,74 @@
+from floodviz.map_utils import site_dict, write_geojson
+
+import unittest
+
+import requests
+import requests_mock
+
+sites = ['05463500', '05471050', '05420680', '05479000', '05484000', '05481000', '05486000', '05421000', '05485500',
```
mbucknell: Test data should be set up in the test class. That's what the setUp function is for. This also guarentees that your test data is the same for each test since it is constructed for each test. If you declare them here, any test could change them, affecting the next test that is run. Alternatively, set up your test data in the individual test that requires it.

Just stylistically, it's not necessary to have this many sites. Testing with a smaller number (say three or four) will make your tests less bulky.

floodviz/map_utils_tests.py
```diff
@@ -0,0 +1,74 @@
...
+
+mock_data = [{'agency_cd': 'USGS', 'site_no': '05411850', 'station_nm': 'Turkey River near Eldorado, IA',
+              'site_tp_cd': 'ST', 'dec_lat_va': '43.0541879', 'dec_long_va': '-91.8090983', 'coord_acy_cd': '1',
+              'dec_coord_datum_cd': 'NAD83', 'alt_va': ' 890.00', 'alt_acy_va': '.01', 'alt_datum_cd': 'NGVD29',
+              'huc_cd': '07060004'}]
+
+
+class TestSiteDict(unittest.TestCase):
+
+    def test_empty_list(self):
```
mbucknell: After refactoring the function, many of these tests will not be necessary. I'll need to review this again once that is done. 

I don't see a test for a good response. I would include a test where the mock request returns good data for a couple of sites.

I would suggest adding a test for missing fields in the dictionary. For instance, mock out a response where huc_cd is missing. You don't need to do all of them.

floodviz/map_utils_tests.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def test_bad_request_data2(self):
+        with requests_mock.Mocker() as m:
+            m.get(url, text='#')
+            self.assertEqual(site_dict(sites, prefix), None)
+
+
+class TestWriteGeojson(unittest.TestCase):
+
+    def test_filename_is_string(self):
```
mbucknell: I'll put this comment here, but you can't really test that the geojson is being formed correctly because this function writes to a file. Maybe a better way to refactor this is to have a function that returns a geojson string (getSitesGeoJson seems like a good name). Then your view code would write that geojson string to a file. And you can add real tests that actually test that a valid geojson string is formed.

floodviz/hydrograph_utils.py
```diff
@@ -66,3 +66,5 @@ def req_hydrodata(sites, start_date, end_date, url_top):
     r = requests.get(url)
     if r.status_code is 200:
         return r.json()['value']['timeSeries']
+    else:
```
skaymen: Fixed.

floodviz/map_utils.py
```diff
@@ -2,15 +2,33 @@
...
     Returns:
-        A dict containing various site information in a usable format.
+        An array of dicts containing various site information in a usable format.
     """
 
+    if not site_list:
+        print("Error: site_list is empty")
+        return
+
+    if not url_prefix:
```
skaymen: Removed this check and associated tests.

floodviz/map_utils.py
skaymen: Removed.

floodviz/map_utils.py
skaymen: Removed.

floodviz/map_utils.py
skaymen: Fixed.

floodviz/map_utils.py
skaymen: Done.

floodviz/map_utils.py
skaymen: Done.

floodviz/map_utils.py
skaymen: Fixed.

floodviz/map_utils.py
skaymen: Fixed.

floodviz/map_utils_tests.py
skaymen: Fixed.

floodviz/map_utils.py
skaymen: Fixed.

floodviz/map_utils_tests.py
skaymen: Done.

floodviz/map_utils_tests.py
skaymen: Done.

PR #19, opened 2017-06-27
D3 mapping

jameeters:


config.py
```diff
@@ -16,3 +16,6 @@
 # This dict holds aspect ratio data for hydrograph rendered in static/js/hydrograph.js. Keys are 'height' and 'width'
 HYDRO_META = {}
 
+
+SPATIAL_REFERENCE_ENDPOINT = 'http://spatialreference.org/ref/epsg/${epsg_code}/proj4/'
+PROJECTION_EPSG_CODE = '2794'
```
mbucknell: I would suggest defaulting this to 4326 which is a global projection. Then override it in instance/config.py. Please update the google doc containing the config.py so we all have easy access to the latest config.py that we are testing with.

floodviz/map_utils.py
```diff
@@ -1,4 +1,7 @@
 import requests
+import json
+from string import Template
```
mbucknell: According to PEP8 standard library imports should be grouped together and put before external package imports. So import requests should be moved below this one.

floodviz/map_utils.py
```diff
@@ -35,24 +38,61 @@ def site_dict(site_list, url_prefix):
...
-           data: the data to be written to the file
+         :param  filename: the file to be written to
+         :param  data: the data to be written to the file
     """
+    data = create_geojson(data)
+    with open(filename, "w+") as file:
+        json.dump(data, file, indent=True)
+
+
+def create_geojson(data):
```
mbucknell: You and @skaymen need to resolve differences in approach here.

floodviz/map_utils.py
```diff
@@ -35,24 +38,61 @@ def site_dict(site_list, url_prefix):
...
-            f.write("\n")
-        f.write(" ] }")
\ No newline at end of file
+
+def projection_info(code, url):
+    """
+    :param code: the EPSG code of the projection we wish to use
+
+    :param url: A python string TEMPLATE in which ${epsg_code} will be replaced with `code`
+        this should found under SPATIAL_REFERENCE_ENDPOINT in `config.py`
```
mbucknell: This part of the sentence is not needed. A utility function should be pretty much standalone. It is not a requirement of this function for this url to found in config.py.

floodviz/views.py
```diff
@@ -18,8 +17,32 @@ def home():
...
     return render_template('index.html')
+
+
+@app.route('/map/')
+def sitemap():
+    site_data = map_utils.site_dict(app.config['SITE_IDS'], app.config['NWIS_SITE_SERVICE_ENDPOINT'])
+    site_data = map_utils.create_geojson(site_data)
+    projection = map_utils.projection_info(app.config['PROJECTION_EPSG_CODE'], app.config['SPATIAL_REFERENCE_ENDPOINT'])
+
+    # TODO: the following is a hack:
```
mbucknell: Why is it a hack?

floodviz/map_utils.py
```diff
@@ -35,24 +38,61 @@ def site_dict(site_list, url_prefix):
...
+
+    :return: the proj4 projection definition string of the desired projection
+    """
+    url = Template(url)
+    url = url.substitute(epsg_code=str(code))
+    req = requests.get(url)
+    assert (req.status_code == 200),\
```
mbucknell: Consider returning an empty string or None or be explicit in the function comments that this function will return an AssertionError if it can't retrieve the project definition. AssertionError seems a little strange to raise in this case.

floodviz/views.py
```diff
@@ -18,8 +17,32 @@ def home():
...
-    
-    with open('floodviz/static/data/hydrograph_data.json', 'w') as fout: 
+
+    with open('floodviz/static/data/hydrograph_data.json', 'w') as fout:
         json.dump(all_series_data, fout, indent=1)
 
     return render_template('index.html')
+
+
+@app.route('/map/')
```
mbucknell: Eventually this should go on the home page but we can take care of that in a subsequent pull request once we determine which way we want to go with the map stuff.

floodviz/views.py
mbucknell: Where did the counties.json file come from? Right now I believe we only need to draw state boundaries but it would be good to clarify with the Read and/or Walker. It does not belong in the instance directory at any rate. So  that explains the comment.

floodviz/views.py
```diff
@@ -18,8 +17,32 @@ def home():
...
+    with open('instance/counties.json', 'r') as bg_file:
+        bg_data = json.load(bg_file)
+    # End hack
+
+
+    mapinfo = {
+        'proj4string': projection,
+        'width': 1000,
+        'height': 600,
+        'scale': .90,  # larger than .97 cuts off data
```
mbucknell: Is this comment true just for this projection? is this dependent on the bounding box or width/height? Also remember that the aspect ratio and width will be specified in config.py and should not be hard coded.

floodviz/templates/sitemap.html
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <link type="text/css" href="{{ url_for('static', filename='bower_components/bootstrap/dist/css/bootstrap.css') }}"/>
+    <title>Site Map</title>
+</head>
+<body>
```
mbucknell: It would be better to have a div specifically for the map. I am fine with this for now as this is more of a proof of concept.

floodviz/static/js/gage_map.js
```diff
@@ -0,0 +1,86 @@
+document.addEventListener("DOMContentLoaded", function (event) {
```
mbucknell: Consider adding JSHint directives such as /* global proj */ . Besides making an IDE using a JSHint plugin happy, it's clearer what this module depends on.

floodviz/static/js/gage_map.js
```diff
@@ -0,0 +1,86 @@
...
+    //Create SVG element
+    var svg = d3.select("body")
+        .append("svg")
+        .attr("width", width)
+        .attr("height", height);
+
+    // read in geojson for sites
+    var site_data = FV.mapinfo.site_data;
+
+    // Calculate bounding box transforms for the sites, we don't care about cutting off the background
```
mbucknell: Actually the bounding box is also passed in via config.py, so this will need to be changed.

floodviz/static/js/gage_map.js
```diff
@@ -0,0 +1,86 @@
...
+    var t = [(width - s * (b[1][0] + b[0][0])) / 2, (height - s * (b[1][1] + b[0][1])) / 2];
+
+    // Update the projection
+    projection
+        .scale(s)
+        .translate(t);
+
+    // Define order in which layers should be added
+    var bg = svg.append("g"),
+        sites = svg.append("g");
```
mbucknell: Each var should have var proceeding it. IMHO it makes it clearer that sites is a variable and not a continuation of the previous line.

floodviz/views.py
jameeters: @skaymen found counties.json. I've been using it because it's technically almost identical to adding state boundaries. I also tried using state borders, but unless I set the scale to about 60 or less, the state lines only showed up around the edges of the map.  I stuck both files in `instance` because I didn't think they should be added to git.

config.py
```diff
@@ -16,3 +15,39 @@
...
+# Map configuration
+
+SPATIAL_REFERENCE_ENDPOINT = 'http://spatialreference.org/ref/epsg/${epsg_code}/proj4/'
+PROJECTION_EPSG_CODE = '4326'
+
+#
+BOUNDING_BOX = [-94.294625, 41.015733, -91.424186, 43.102420]
+
+BACKGROUND_FILE = 'floodviz/static/geojson/states.json'
+REFERENCE_FILE = 'floodviz/static/geojson/reference.json'
```
mbucknell: What is in reference.json? Is this the features that the data scientist is providing? If so, it has to be text in the instance/config.py.

config.py
```diff
@@ -16,3 +15,39 @@
...
+# this contains data that needs no further transformation before being sent to the javascript
+MAP_CONFIG = {
```
mbucknell: I would recommend that the map_config include the width, height, and scale. If you want the bounding box represented as a geojson, then do so in the view or in the application initialization.

floodviz/map_utils.py
```diff
@@ -52,26 +55,61 @@ def site_dict(site_list, url_prefix):
...
+
+    :param url: A python string TEMPLATE in which ${epsg_code} will be replaced with `code`
+        A valid example is SPATIAL_REFERENCE_ENDPOINT in `config.py`
+
+    :return: the proj4 projection definition string of the desired projection, or None if no such projection is found
+    """
+    url = Template(url)
+    url = url.substitute(epsg_code=str(code))
+    req = requests.get(url)
+    if req.status_code == '404':
```
mbucknell: I would actually check for a 200. The status_code could also be 500.

floodviz/static/geojson/reference.json
```diff
@@ -0,0 +1,31 @@
+{
```
mbucknell: See previous comment. This is configuration information and should not be part of the repo.

floodviz/views.py
```diff
@@ -1,13 +1,32 @@
...
 
 
+@app.route('/')
+def root():
+    _hydrograph_helper()
+    mapinfo = _map_helper()
+    return render_template('index.html', mapinfo=mapinfo)
+
+
 @app.route('/home/')
```
mbucknell: Let's call this one '/hydrograph/'

floodviz/views.py
```diff
@@ -1,13 +1,32 @@
...
 def home():
+    _hydrograph_helper()
+    return render_template('hydrograph.html')
+
+
+@app.route('/map/')
+def sitemap():
```
mbucknell: This will help accomplish the goal of having embeddable web visuals so great idea!

floodviz/map_utils.py
```diff
@@ -52,26 +55,61 @@ def site_dict(site_list, url_prefix):
     return dnice
 
 
-def get_sites_geojson(data):
-    """Writes site data to a .json file so it can be mapped
-       Args:
-           data: A list of dicts with data to be written to the file
-        Returns:
-            A string containing geojson data
+def write_geojson(filename, data):
```
mbucknell: Is this still used? If not, it should be removed.

floodviz/templates/index.html
```diff
@@ -14,19 +14,23 @@
...
 		var FV = FV || {};
 		FV.hydrometa = {{ config.HYDRO_META|safe }};
+		FV.mapinfo = {{ mapinfo | tojson |safe }};
 	</script>
 
 	<script type="text/javascript" src="{{ url_for('static', filename='bower_components/jquery/dist/jquery.min.js') }}"></script>
 	<script type="text/javascript" src="{{ url_for('static', filename='bower_components/bootstrap/dist/js/bootstrap.js') }}"></script>
 	<script type="text/javascript" src="{{ url_for('static', filename='bower_components/d3/d3.js') }}"></script>
 	<script type="text/javascript" src="{{ url_for('static', filename='bower_components/nvd3/build/nv.d3.min.js') }}"></script>
+    <script type="text/javascript" src="{{ url_for('static', filename='bower_components/proj4/dist/proj4.js') }}"></script>
```
mbucknell: should use tabs rather than spaces

floodviz/map_utils.py
```diff
@@ -52,26 +55,61 @@ def site_dict(site_list, url_prefix):
...
+                'name': original['station_nm'],
+                'id': original['site_no'],
+                'huc': original['huc_cd']
+            }
+        }
+        new_data['features'].append(item)
+    return new_data
+
+
+def projection_info(code, url):
```
mbucknell: we should add tests for this. Don't won't to hold up merging for that. Add the tests when you have time.

config.py
jameeters: reference.json comes from the example reference locations [attached to FV-11](https://internal.cida.usgs.gov/jira/secure/attachment/21203/flood_cities_01.json). I think it would be better to leave it as a file because it might be very large, but I think moving the file to `instance/` makes sense.

mbucknell: I would prefer for now, it just go directly into config.py. We need to discuss as a team how we really want to read in configuration parameters. For now, I'd like them collected in a single place.

floodviz/map_utils.py
```diff
@@ -1,73 +1,10 @@
 import json
+import xml.etree.ElementTree as ET
```
mbucknell: Remove unused import

floodviz/map_utils.py
```diff
@@ -113,3 +50,48 @@ def projection_info(code, url):
...
+        return []
+
+    # generate the string of site ids for the url
+    id_input_string = ",".join(ids)
+    url = url_prefix + "site/?format=mapper&sites=" + id_input_string + "&siteStatus=all"
+
+    req = requests.get(url)
+    if req.status_code != 200:
+        print('request to NWIS has failed')
+        return []
```
mbucknell: I would prefer for a failure of the request that None is returned. That is what the function did originally. The caller of the function can then distinguish between no data (which could be a legitimate response) and not receiving a response at all.

floodviz/map_utils.py
```diff
@@ -113,3 +50,48 @@ def projection_info(code, url):
...
+    :return: list of dictionaries describing the sites
+    note that huc_cd, along with some other fields, will not be included
+    """
+    if not ids:
+        print('No site IDs provided')
+        return []
+
+    # generate the string of site ids for the url
+    id_input_string = ",".join(ids)
+    url = url_prefix + "site/?format=mapper&sites=" + id_input_string + "&siteStatus=all"
```
mbucknell: I'm going to push back on this. The mapper format is being used for a specific application in the USGS. I'd like to stick to the rdb format as in it well documented and includes complete information.

floodviz/map_utils.py
```diff
@@ -41,37 +45,77 @@ def site_dict(site_list, url_prefix):
         else:
             break
 
-    # make a list of dicts from data
-    fields = data[0].split('\t')
+    keep_fields = [
```
mbucknell: This is nice!

floodviz/map_utils.py
```diff
@@ -41,37 +45,77 @@ def site_dict(site_list, url_prefix):
...
+    fields = [field.strip() for field in data[0].split('\t')]
     dnice = []
     for line in data[2:]:
-        line = line.split('\t')
+        line = [item.strip() for item in line.split('\t')]
         line_dict = dict(zip(fields, line))
-        dnice.append(line_dict)
+        try:
+            filtered_line_dict = {k: line_dict[k] for k in keep_fields}
+        except KeyError:
```
mbucknell: If you are going to raise KeyError, need to reflect that in the function comments. However, I think I would just use line_dict.get(k, '') and default to nothing. Or we could just not return data for that line. It's fine to leave this as is for now, but something to consider.

floodviz/tests/test_map_utils.py
```diff
@@ -1,205 +1,171 @@
...
-        self.missing_huc_return = [{'agency_cd': 'USGS', 'site_no': '05420680',
-                             'station_nm': 'Wapsipinicon River near Tripoli, IA', 'site_tp_cd': 'ST',
-                             'dec_lat_va': '42.83609117', 'dec_long_va': '-92.2574003', 'coord_acy_cd': 'F',
-                             'dec_coord_datum_cd': 'NAD83', 'alt_va': ' 986.42', 'alt_acy_va': '.01',
-                             'alt_datum_cd': 'NGVD29', 'huc_cd': '07080102'},
-                            {'agency_cd': 'USGS', 'site_no': '05463500', 'station_nm': 'Black Hawk Creek at Hudson, IA',
-                             'site_tp_cd': 'ST', 'dec_lat_va': '42.4077639', 'dec_long_va': '-92.4632451',
-                             'coord_acy_cd': 'F', 'dec_coord_datum_cd': 'NAD83', 'alt_va': ' 865.03',
-                             'alt_acy_va': '.01', 'alt_datum_cd': 'NGVD29'}]
+        self.prefix = 'https://waterservices.usgs.gov/nwis/'
```
mbucknell: These should not use the real services. Real services should never be used for unit tests. Even though we are mocking, it's possible that we change our code in way that the real service would get hit for a unit test. This has to be changed back. I'm going to merge anyways, but please fix this.

PR #21, opened 2017-06-27
Unit Testing for Hydrograph_utils

ede0m:
Dockerfile nosetests are not working. This also has problems with the docker file for some reason. The server executed by docker run will always 404 the /home/ page.


floodviz/test/hydrograph_utils_tests.py
```diff
@@ -0,0 +1,74 @@
...
+    self.assertEqual(req_hydrodata(['x'], self.H_START_DT, self.H_END_DT, self.prefix), [])
+    self.assertEqual(req_hydrodata([123456, 100], self.H_START_DT, self.H_END_DT, self.prefix), [])
+    self.assertEqual(req_hydrodata([], self.H_START_DT, self.H_END_DT, self.prefix), [])
+
+  def test_bad_url(self):
+    self.assertEqual(req_hydrodata(self.sites, self.H_START_DT, self.H_END_DT, 'badurl.noob' ), [])
+    self.assertEqual(req_hydrodata(self.sites, self.H_START_DT, self.H_END_DT, ''), [])
+
+  def test_valid_data(self):
+    self.assertTrue(req_hydrodata(self.sites, self.H_START_DT, self.H_END_DT, self.prefix))
```
mbucknell: So these tests will probably go away with the changes I've asked for. You should look at using requests-mock to mock the requests.get. It's fairly simple to use. Then you can test receiving good data and test for receiving an non 200 status code.  I would like a test for passing in an empty list as the data and checking to see that an empty list is returned as well.

floodviz/test/hydrograph_utils_tests.py
```diff
@@ -0,0 +1,74 @@
...
+
+  def test_data_not_list(self):
+    self.assertEqual(parse_hydrodata("Not a REAL list"), [])
+    self.assertEqual(parse_hydrodata({}), [])
+
+  def invalid_elements(self):
+    self.assertEqual(parse_hydrodata([{} ,'A', 'B']), [])
+
+  def test_valid_data(self):
+    self.assertTrue(parse_hydrodata(self.mock_data))
```
mbucknell: You should be checking what the function returns not just that it returns something

floodviz/test/hydrograph_utils_tests.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+
+if __name__ == '__main__':
```
mbucknell: This is not really necessary if we are using nosetests to run our test suites. You can leave it unless it complicates running other things.

floodviz/hydrograph_utils.py
```diff
@@ -2,12 +2,12 @@
...
 import requests
 
-
 def parse_hydrodata(jdata):
     
     """ 
     Parses json Hydrodata from NWIS webservice
-    and formats for NVD3 charting library.
+    and formats for NVD3 charting library. Upon failure, this will
+    return empty data list.
```
ede0m: I choose to use an empty list for fault returns because an empty list sparks a nice message within NVD3 that shows 'No data Available'. None type does not do this.

floodviz/hydrograph_utils.py
```diff
@@ -48,7 +54,8 @@ def parse_hydrodata(jdata):
 def req_hydrodata(sites, start_date, end_date, url_top):
 
     """ 
-    Requests hydrodata from nwis web service based on passed in parameters. 
+    Requests hydrodata from nwis web service based on passed in parameters.
+    Upon request failure, this will return an empty data list.
```
ede0m: None should be fine for this return value.

floodviz/hydrograph_utils.py
```diff
@@ -60,9 +67,21 @@ def req_hydrodata(sites, start_date, end_date, url_top):
         returns a dictonary with the requested data from the nwis service
```
ede0m: Technically, yes. However since the string.join(siteIDs), the list will always contain just one dict with all series values. It just has to be wrapped in a list for NVD3. I suppose I could do this elsewhere.

floodviz/hydrograph_utils.py
mbucknell: I think that's fine then. We can always reconsider later if this function is used for something else.

floodviz/hydrograph_utils.py
mbucknell: Okay, this may warrant a little more explanation then so the user of the function knows what is returned. But it is a list that is returned even if it's a single element.

Dockerfile
```diff
@@ -26,6 +26,8 @@ RUN npm config set cafile "/app/DOIRootCA2.cer" \
 
 RUN node_modules/bower/bin/bower install --allow-root
 
+#RUN nosetests --all-modules --exe
```
ede0m: fixed

floodviz/hydrograph_utils.py
```diff
@@ -19,10 +19,16 @@ def parse_hydrodata(jdata):
...
+    
+    if not isinstance(jdata, list):
+        return all_series_data
+    
     for idx, site in enumerate(jdata):
+        # Check that list elements are dicts
+        if not isinstance(site, dict):
+            return all_series_data
+
         site_name = site['sourceInfo']['siteName']
```
ede0m: What about try/except KeyError? Or is .get preferred?

mbucknell: This is one of those religious wars. Should there be a bunch of try except handlers or should you make an attempt to handle a dictionary whose format may not be what you expect? I think it's clearer to do the later and IMHO the code is cleaner. We've actually talked about this issue in my scrum team not to long ago.

floodviz/test/hydrograph_utils_tests.py
ede0m: As for the last test you mentioned (passing in an empty list and expecting a returned empty list), wouldn't my test_empty_data in TestParseHydroData work for this?

mbucknell: Yes. Sorry must have missed it.

PR #22, opened 2017-06-28
Major refactoring to maputils functions and unit tests

skaymen:
I have implemented the suggested changes to my functions as well as created more sensible unit tests.


mbucknell:
Going ahead and merging so that James will have an easier time making updates to his code. I'll still be reviewing it later.

Dockerfile
```diff
@@ -26,6 +26,10 @@ RUN npm config set cafile "/app/DOIRootCA2.cer" \
 
 RUN node_modules/bower/bin/bower install --allow-root
 
+RUN nosetests --all-modules --exe
+
+RUN ls *
```
mbucknell: What is the purpose of this? I suspect it's for debugging and should probably be removed.

Dockerfile
```diff
@@ -26,6 +26,10 @@ RUN npm config set cafile "/app/DOIRootCA2.cer" \
 
 RUN node_modules/bower/bin/bower install --allow-root
 
+RUN nosetests --all-modules --exe
```
mbucknell: Just reading about nosetests. Why are you using --exe? I haven't had to use this in the past.

Dockerfile
mbucknell: Is this for debug? If so it should be removed.

floodviz/map_utils_tests.py
```diff
@@ -0,0 +1,171 @@
+from floodviz.map_utils import site_dict, get_sites_geojson
```
mbucknell: Should have said something early. It's our practice here to name put our tests in a directory and call the file test_*. So please move this to floodviz/tests/test_map_utils.py

floodviz/map_utils_tests.py
```diff
@@ -0,0 +1,171 @@
...
+        self.assertEqual(site_dict(self.sites, self.prefix), self.good_return)
+
+    def test_missing_fields(self):
+        with requests_mock.Mocker() as m:
+            m.get(self.url, text=self.mock_missing_huc)
+            self.assertEqual(site_dict(self.sites, self.prefix), self.missing_huc_return)
+
+    def test_bad_status_code(self):
+        with requests_mock.Mocker() as m:
+            m.get(self.url, status_code=201)
```
mbucknell: It would be better to test this with a status_code which represents an error. A 201 actually is a success status code, see https://en.wikipedia.org/wiki/List_of_HTTP_status_codes. This particular service does not use that status_code. It would be preferable to test this using a 404 or 500.

floodviz/map_utils_tests.py
```diff
@@ -0,0 +1,171 @@
...
+                                   " }\n"
+                                   " ] }")
+
+    def test_good_data(self):
+        self.assertEqual(get_sites_geojson(self.good_data), self.good_return)
+
+    def test_missing_huc(self):
+        self.assertEqual(get_sites_geojson(self.missing_huc_data), self.missing_huc_return)
+
+if __name__ == '__main__':
```
mbucknell: This is not necessary when using nose to run your tests. The reason to use nose is it collects all of your tests for you. So you can run the tests for an entire application with one command.

Dockerfile
skaymen: The tests did not run during the Docker build unless I used --exe. I'm not sure if there is another way around this. I can look into it.

Dockerfile
skaymen: Good catch, thanks.

skaymen: Removed.

floodviz/map_utils_tests.py
skaymen: Fixed.

floodviz/map_utils_tests.py
skaymen: Removed.

Dockerfile
mbucknell: I took it out and did a Docker build and the tests did run. Maybe try again without it?

floodviz/map_utils_tests.py
skaymen: Fixed.

Dockerfile
skaymen: Still having the same issue-- the build is successful, but it runs 0 tests. I'll do some googling.

skaymen: Garritt is having this issue as well.

mbucknell: I wonder if this is a Windows issue. I know Andrew Yan has run nosetest on his Windows machine without having to use that flag. It might be worthwhile pinging him.

skaymen: Ok, I'll send him a message.

PR #24, opened 2017-06-28
Lollipop - Bare D3 lollichart. 

ede0m:
Reverted to using bare d3 because of [this issue](https://stackoverflow.com/questions/44806887/dynamic-cx-cy-svg-circle-coordinates). I still have to refine the chart titles and placement and what not.


floodviz/peak_flow_utils.py
```diff
@@ -0,0 +1,53 @@
...
+        end_date - ending date to chart peak flow data
+        url_peak - config varbile for nwis waterdata service endpoint url
+    
+    RETURNS:
+        content - list of all lines in the data file 
+
+    """
+    url = url_peak + '?site_no=' + site + '&agency_cd=USGS&format=rdb'
+    r = requests.get(url)
+    if r.status_code is 200:
```
ede0m: Fixed to return an empty List

PR #27, opened 2017-06-29
Voronoi

skaymen:
Chart now uses a Voronoi tesselation and prints the timezone offset in addition to time and date.

### DIFF
--- a/floodviz/hydrograph_utils.py
+++ b/floodviz/hydrograph_utils.py
@@ -24,7 +24,8 @@ def parse_hydrodata(jdata):
 
     for idx, site in enumerate(jdata):
         site_name = site['sourceInfo']['siteName']
-        all_series_data.append({'key': site_name, 'values': [], 'max_val': 0})
+        site_id = site['sourceInfo']['siteCode'][0]['value']
+        all_series_data.append({'key': site_id, 'values': []})
 
         # Fill new data for this series
         for idx2, obj in enumerate(site['values'][0]['value']):
@@ -38,7 +39,7 @@ def parse_hydrodata(jdata):
             dt_ms = time.mktime(dt.timetuple()) * 1000
             # (for below if statment) create dummy value for nvd3 issue at https://github.com/novus/nvd3/issues/695 #
             if idx2 is 0: # First datapoint of this site
-                all_series_data[idx]['values'].append({'date': date, "time": 0, "time_mili": dt_ms, 'value': 0})
+                all_series_data[idx]['values'].append({'site_name': site_name, 'date': date, "time": 0, "time_mili": dt_ms, 'value': 0})
 
             all_series_data[idx]['values'].append({'date': date, "time": t, "time_mili": dt_ms, 'value': value})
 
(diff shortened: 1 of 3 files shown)


floodviz/static/js/hydrograph.js
```diff
@@ -18,25 +18,27 @@ document.addEventListener("DOMContentLoaded", function(event) {
...
 					.showLegend(false)
 					.showControls(false);
-
-			chart.xAxis.axisLabel(" Date (M-D-Y)")
+			
+			chart.xAxis.axisLabel(" Date (M-D-Y H:M:S)")
 					.axisLabelDistance(10)
 					.ticks(5)
-					.tickFormat(function(d){return d3.time.format('%m-%d-%y')(new Date(d))});
+					.tickFormat(function(d){return d3.time.format('%m-%d-%y %H:%M:%S %Z')(new Date(d))});
```
ede0m: Do we want to display the Hours/Seconds if we will be plotting over many days?

floodviz/hydrograph_utils.py
```diff
@@ -38,9 +39,11 @@ def parse_hydrodata(jdata):
             dt_ms = time.mktime(dt.timetuple()) * 1000
             # (for below if statment) create dummy value for nvd3 issue at https://github.com/novus/nvd3/issues/695 #
             if idx2 is 0: # First datapoint of this site
-                all_series_data[idx]['values'].append({'date': date, "time": 0, "time_mili": dt_ms, 'value': 0})
+                all_series_data[idx]['values'].append({'site_name': site_name, 'date': date, "time": 0,
```
ede0m: Should we maybe have sitename as another key on line 28 instead of having it recorded for every datapoint?

floodviz/hydrograph_utils.py
```diff
@@ -24,7 +24,8 @@ def parse_hydrodata(jdata):
 
     for idx, site in enumerate(jdata):
         site_name = site['sourceInfo']['siteName']
-        all_series_data.append({'key': site_name, 'values': [], 'max_val': 0})
+        site_id = site['sourceInfo']['siteCode'][0]['value']
+        all_series_data.append({'key': site_id , 'values': []})
```
ede0m: I agree that having the main key as the site id is more useful; however, maybe add another key for the site name here instead of for every data point.

floodviz/static/js/hydrograph.js
mbucknell: I think not, but it is a question for the product owner.

skaymen: The ticket does ask for time and timezone in the tooltip, but I can also remove this if that's better. Maybe make it so time is in tooltip, but only date is on x-axis?

mbucknell: Yes, I looked at the ticket and Stan's right. They likely want Hours/Seconds. This ticket doesn't address what is on the x-axis.

floodviz/hydrograph_utils.py
skaymen: Made key a list containing both items.

floodviz/hydrograph_utils.py
skaymen: Changed.

floodviz/hydrograph_utils.py
skaymen: Actually putting name as a separate key since ticket specifies site ID only on tooltip.

PR #28, opened 2017-06-29
Changed unit tests to use a fake url

skaymen:
Altered tests so that they can be run without querying a real website or access to the Internet.

### DIFF
--- a/floodviz/tests/test_map_utils.py
+++ b/floodviz/tests/test_map_utils.py
@@ -9,8 +9,8 @@ class TestSiteDict(unittest.TestCase):
 
     def setUp(self):
         self.sites = ['05420680', '05463500']
-        self.prefix = 'https://waterservices.usgs.gov/nwis/'
-        self.url = "https://waterservices.usgs.gov/nwis/site/?format=rdb&sites=05420680,05463500&siteStatus=all"
+        self.mock_prefix = 'https://fake.fake/nwis/'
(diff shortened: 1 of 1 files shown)


floodviz/tests/test_map_utils.py
```diff
@@ -9,8 +9,8 @@ class TestSiteDict(unittest.TestCase):
 
     def setUp(self):
         self.sites = ['05420680', '05463500']
-        self.prefix = 'https://waterservices.usgs.gov/nwis/'
-        self.url = "https://waterservices.usgs.gov/nwis/site/?format=rdb&sites=05420680,05463500&siteStatus=all"
+        self.mock_prefix = 'https://fake.fake/nwis/'
+        self.mock_url = "https://fake.fake/nwis/site/?format=rdb&sites=05420680,05463500&siteStatus=all"
```
mbucknell: This isn't DRY. Use self.mock_prefix to form self.mock_url

floodviz/tests/test_map_utils.py
```diff
@@ -68,20 +105,24 @@ def setUp(self):
                              'alt_acy_va': '.01', 'alt_datum_cd': 'NGVD29'}]
 
     def test_empty_list(self):
-        self.assertEqual(site_dict([], self.prefix), [])
+        with requests_mock.Mocker() as m:
```
mbucknell: This isn't really necessary but harmless. If you had a more general url matching it might make more sense (e.g. any url with self.mock_prefix. I'm okay with leaving it.

PR #30, opened 2017-07-03
D3 mapping

jameeters:
Add function to filter background file to only the necessary features, and switch gage points to svg circles instead of paths.

### DIFF
--- a/floodviz/map_utils.py
+++ b/floodviz/map_utils.py
@@ -124,14 +124,37 @@ def filter_background(bbox, bg_filename):
     features = bg['features']
     in_box = []
     for f in features:
+        added = False
         coordinates = f['geometry']['coordinates']
+        feature_bbox = {
+            'min': {
+                'lat': float('inf'),
+                'lon': float('inf')
+            },
+            'max': {
+                'lat': float('-inf'),
+                'lon': float('-inf')
+            }
+        }
         for group in coordinates:
             if f['geometry']['type'] == 'MultiPolygon':
                 group = group[0]
             for pair in group:
+                feature_bbox['min']['lat'] = min(feature_bbox['min']['lat'], pair[0])
+                feature_bbox['min']['lon'] = min(feature_bbox['min']['lon'], pair[1])
+                feature_bbox['max']['lat'] = max(feature_bbox['max']['lat'], pair[0])
+                feature_bbox['max']['lon'] = max(feature_bbox['max']['lon'], pair[1])
                 if min(box_lat) <= pair[0] <= max(box_lat) and min(box_lon) <= pair[1] <= max(box_lon):
                     in_box.append(f)
+                    added = True
                     break
+            if not added and \
+                    feature_bbox['min']['lat'] < min(box_lat) and \
+                    feature_bbox['max']['lat'] > max(box_lat) and \
+                    feature_bbox['min']['lon'] < min(box_lon) and \
+                    feature_bbox['max']['lon'] > max(box_lon):
+                in_box.append(f)
+                added = True
 
     keepers = {
         'type': 'FeatureCollection',
--- a/floodviz/map_utils.py
+++ b/floodviz/map_utils.py
@@ -115,7 +115,26 @@ def projection_info(code, url):
     return req.text
 
 
+def filter_background(bbox, bg_filename):
+    box_lat = [bbox[0], bbox[2]]
+    box_lon = [bbox[1], bbox[3]]
 
+    with open(bg_filename, 'r') as bg_file:
+        bg = json.load(bg_file)
+    features = bg['features']
+    in_box = []
+    for f in features:
+        coordinates = f['geometry']['coordinates']
+        for group in coordinates:
+            if f['geometry']['type'] == 'MultiPolygon':
+                group = group[0]
+            for pair in group:
+                if min(box_lat) <= pair[0] <= max(box_lat) and min(box_lon) <= pair[1] <= max(box_lon):
+                    in_box.append(f)
+                    break
 
-
-
+    keepers = {
+        'type': 'FeatureCollection',
+        'features': in_box
+    }
+    return keepers
--- a/floodviz/map_utils.py
+++ b/floodviz/map_utils.py
@@ -116,48 +116,61 @@ def projection_info(code, url):
 
 
 def filter_background(bbox, bg_filename):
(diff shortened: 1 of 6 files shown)


skaymen:
I tested the code and verified it works. I do not have any issues with the code as it is written currently.

floodviz/map_utils.py
```diff
@@ -115,7 +115,62 @@ def projection_info(code, url):
...
+        
+        coordinates = f['geometry']['coordinates']
+
+        for group in coordinates:
+            if len(in_box) > starting_len:
+                # This feature has already been added
+                break
+            # actual points for MultiPolygons are nested one layer deeper than those for polygons
+            if f['geometry']['type'] == 'MultiPolygon':
+                group = group[0]
```
mbucknell: Rather than reassigning group[0] to group, declare a new variable (called maybe geometry?). Also does this only get the first polygon of coordinates from a multipolygon?

floodviz/map_utils.py
```diff
@@ -115,7 +115,62 @@ def projection_info(code, url):
...
+
+            for pair in group:
```
mbucknell: Consider changing to:  for lon, lat in group

floodviz/map_utils.py
```diff
@@ -115,7 +115,62 @@ def projection_info(code, url):
...
+
+                # We only need to check the box around the state if we don't add the state based on an intersection.
+                feature_min_lon = min(feature_min_lon, pair[0])
+                feature_min_lat = min(feature_min_lat, pair[1])
+                feature_max_lon = max(feature_max_lon, pair[0])
+                feature_max_lat = max(feature_max_lat, pair[1])
+
+        # If the box containing a feature also contains the bounding box, keep this feature
+        # Allow adding more than one because otherwise MD contains boxes in WV, and CA would contain most of NV.
+        if feature_min_lat < min(box_lat) and feature_max_lat > max(box_lat) and \
```
mbucknell: I am a little confused. I would think that you would want to keep a feature if it is entirely within the bbox not if it is entirely outside the bbox?

floodviz/static/css/mapstyle.css
```diff
@@ -0,0 +1,11 @@
+circle.gage-point{
+    stroke: blue;
+    stroke-width: 2;
+    fill-opacity: 0;
+}
+
+circle.debug-point{
```
mbucknell: If this is for debug, it should be removed.

floodviz/static/js/gage_map.js
```diff
@@ -101,12 +103,15 @@ document.addEventListener("DOMContentLoaded", function (event) {
         var bbox = svg.append("g");
         var bbox_data = FV.mapinfo.bounds;
         // Add geojson to map
-        bbox.selectAll("path")
-            .data(bbox_data.features)
-            .enter()
-            .append("path")
-            .attr("d", path)
-            .style("fill", "green");
+        bbox.selectAll("circle")
```
mbucknell: If this is for debug, it should be removed.

floodviz/map_utils.py
jameeters: This keeps any feature that entirely contains the bbox. If the bbox falls entirely within a state, the state won't have any borders within the bbox and so wouldn't be kept without this.

mbucknell: But wouldn't this mean that we wouldn't see it at all anyways?

mbucknell: I would also think, we want to have a state that falls entirely within the bounding box regardless of the answer to the above question.

jameeters: Not necessarily, because if the aspect ration isn't 1:1 we end up showing areas outside the bounding box. 

I think my comments in the code are a bit confusing. Any feature with that has a point within the bounding box is kept, so we do keep any feature that falls entirely inside the bounding box.

floodviz/map_utils.py
jameeters: This is because the lat-lon pairs for multipolygons are nested a layer deeper than they are in polygons.
Points in polygons are formatted as 
```
[
    [
        [lon, lat],
        [lon, lat]
    ]
] 
```
But in multipolygons they are formatted as 
```
[
    [[
        [lon, lat],
        [lon, lat]
    ]], 
    [[
        [lon, lat],
        [lon, lat]
    ]]
] 
```
so for multiploygons, `group` is a list containing one list: `group[0]` which contains the points.

floodviz/static/css/mapstyle.css
jameeters: It is for debug, but it's really nice to be able to see the bounding box on the map, and I'd rather not have to add and remove it all the time. It's disabled by default and doesn't do anything unless you add `debug = True` in `MAP_CONFIG`

jameeters:
Closing to make requested changes.

floodviz/map_utils.py
mbucknell: It's bad practice to reassign the value of a loop variable. I understand why you have to do it, but would prefer that a new variable is created with the array of lon, lats.

floodviz/map_utils.py
mbucknell: 👍

floodviz/static/css/mapstyle.css
mbucknell: If it's tied to DEBUG (the config variable) then I'm okay with leaving it in.

floodviz/map_utils.py
jameeters: I'd prefer that too, but then I would end up with
```
if multipolygon:
    for pair in group[0]:
         . . .
else:
    for pair in group:
         . . .
```
And I think that's worse.

I could look at factoring out the contents of the loop into a new function, but I think I tried that earlier and gave up because those operations depend on too many variables set in the outer loops.

floodviz/static/css/mapstyle.css
jameeters: It isn't now, but it easily could be. I'll change it.

floodviz/map_utils.py
mbucknell: Could you change it to: ```
if multipolgyon:
    geom = group[0]
else:
    geom = group
for lon, lat in geom
```

jameeters: Ok, I'll change it to that.

PR #36, opened 2017-07-06
D3 mapping

jameeters:


floodviz/tests/test_map_utils.py
```diff
@@ -1,10 +1,9 @@
...
-from floodviz.map_utils import site_dict, create_geojson, projection_info
-
 import unittest
 
 import requests_mock
+from nose.tools import raises
+
+from floodviz.map_utils import site_dict, create_geojson, projection_info, filter_background
 
 LIST_OF_SITE_DICTS = [
```
mbucknell: This should be defined inside TestSiteDict. All test data should be defined within the test class. Not sure when this happened but it should be fixed at some point.

PR #38, opened 2017-07-06
Reference rivers

jameeters:


floodviz/map_utils.py
```diff
@@ -1,3 +1,5 @@
+import json
```
skaymen: May be unnecessary import.

config.py
```diff
@@ -41,7 +41,8 @@
 
 # paths to geojson files for background and reference locations
 BACKGROUND_FILE = ''
-REFERENCE_FILE = ''
+RIVERS_FILE = ''
```
mbucknell: Okay for this, but we will need to include the rivers in the REFERENCE_FILE

floodviz/static/js/map.js
```diff
@@ -0,0 +1,106 @@
...
+    add_paths(FV.mapinfo.bg_data, "background");
+
+    add_paths(FV.mapinfo.rivers_data, "river");
+
+    add_circles(FV.mapinfo.ref_data, "ref-point", 2);
+
+    add_circles(FV.mapinfo.site_data, "gage-point", 3);
+
+
+    if (FV.mapinfo.debug) {
```
mbucknell: I don't see that debug is getting set in mapinfo. Consider putting the config DEBUG value in FV. So in index.html it would be FV = {debug: config.DEBUG}

floodviz/static/js/map.js
```diff
@@ -0,0 +1,106 @@
...
+        add_circles(FV.mapinfo.bounds, "debug-point", 3)
+    }
+    /**
```
mbucknell: Consider grouping all helper functions together. These two are at the bottom of the file, the degreesToRadians and radiansToDegrees are at the top. I

floodviz/static/js/map.js
```diff
@@ -0,0 +1,106 @@
...
+     * Add circles to the map.
+     * @param data The geojson to be added to the svg
```
mbucknell: Should indicate the data type. For instance is radius a number or a string (or can it be either).

PR #39, opened 2017-07-07
Map tooltip

ede0m:


floodviz/static/css/mapstyle.css
```diff
@@ -8,4 +8,23 @@ circle.debug-point{
...
     stroke-width: 2;
     fill-opacity: 0;
+}
+
+.maptip {
+	position: absolute;
+  	display: none;
+  	min-width: 80px;
+  	height: auto;
+  	background: none repeat scroll 0 0 #ffffff;
```
jameeters: This should be `background-color`

floodviz/static/js/gage_map.js
```diff
@@ -1,118 +1,133 @@
...
+		.data(site_data.features)
+		.enter()
+		.append("circle")
+		.attr("r", 3)
+		.attr("transform", function (d) {
+			return "translate(" + projection(d.geometry.coordinates) + ")";
+		})
+		.attr("class", "gage-point")
+		.on('mousemove', function(d) {
+			maptip.transition().duration(500).style("opacity", .9);
```
jameeters: It's better to set the opacity in CSS.

PR #42, opened 2017-07-07
Fixed Tests

skaymen:
Implemented suggested changes and fixed tests.

### DIFF
--- a/floodviz/templates/index.html
+++ b/floodviz/templates/index.html
@@ -24,7 +24,7 @@
 	<script type="text/javascript">
 		// Global Namespace for project - FV
 		var FV = FV || {};
-		FV.hydrometa = {{ config.HYDRO_META | safe }};
+		FV.chartdims = {{ config.CHART_DIMENSIONS | safe }}
 		FV.peakmeta = {{ config.PEAK_META | safe }}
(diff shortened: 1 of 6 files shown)


floodviz/templates/index.html
```diff
@@ -24,7 +24,7 @@
 	<script type="text/javascript">
 		// Global Namespace for project - FV
 		var FV = FV || {};
-		FV.hydrometa = {{ config.HYDRO_META | safe }};
+		FV.chartdims = {{ config.CHART_DIMENSIONS | safe }}
```
mbucknell: These all should end with a semicolon.

floodviz/tests/test_hydrograph_utils.py
```diff
@@ -109,5 +109,11 @@ def test_data_not_list(self):
         self.assertEqual(parse_hydrodata({}), [])
 
     def test_valid_data(self):
-        ret = parse_hydrodata(self.mock_data)
-        self.assertEqual(ret, self.mock_parsed_data)
+            # Delete 'time_mili' key before comparing due to python time conversion inconsistencies.
```
mbucknell: You aren't actually deleting the time_mili key so this comment should be removed. I think it would be good to add a test to make sure that parse_hydrodata returns the correct number of list elements too.

PR #43, opened 2017-07-10
Frozen filepath

ede0m:
There is a warning when running docker image. 
image

Most likely because content-type is 'application/json' but this view is type 'text/html'


Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN npm config set cafile "/app/DOIRootCA2.cer" \
 
 RUN node_modules/bower/bin/bower install --allow-root
 
-RUN nosetests --all-modules --exe
+#RUN nosetests --all-modules --exe
```
mbucknell: I think this commented out because of the failing test. I'll fix this once Stan's changes are merged.

floodviz/templates/index.html
```diff
@@ -3,7 +3,8 @@
 
 <head>
 	<meta charset="UTF-8">
-	<link rel="stylesheet" type="text/css" href="{{ url_for('static', filename='bower_components/bootstrap/dist/css/bootstrap.css') }}"/>
+	<link rel="stylesheet" type="text/css"
```
mbucknell: Stop changing this back and forth. This particular line has been change with almost every pull request.

PR #49, opened 2017-07-11
Link map to timeseries

jameeters:
Emphasize hydrograph lines when mousing over the gage on the map.


floodviz/static/js/map.js
```diff
@@ -66,14 +66,18 @@ document.addEventListener("DOMContentLoaded", function (event) {
 	add_circles(FV.mapinfo.ref_data, "ref-point", 2);
 	var sites = add_circles(FV.mapinfo.site_data, "gage-point", 3);
 	sites.selectAll("circle")
-		.on('mousemove', function (d) {
+		.on("mouseover", function (d) {
```
skaymen: This could be its own function, as in hydrograph.js

floodviz/static/css/style.css
```diff
@@ -16,3 +16,4 @@ body {
 	stroke-width: 1;
 }
 
+
```
skaymen: Extra line added at bottom of file.

floodviz/static/js/map.js
jameeters: I think it's better as an anonymous function because we may end up adding different mouseover actions to the other features on the map.

PR #53, opened 2017-07-11
Config examples

jameeters:


examples/iowa.py
```diff
@@ -0,0 +1,89 @@
...
+        }
+    ]
+}
+
+### Peak Flow Config Vars ###
+
+# Site ID for peak flow bar chart
+PEAK_SITE = '05481950'  # used arbitrary site for now --- 05458900
+# start date for peak flow values
+PEAK_START_DT = '2008-05-20'  ## THIS VALUE REALLY ISN'T USED
```
mbucknell: If it's not used, let's remove it from the code. I believe that PEAK_END_DT really shouldn't be used either. It's the event end date. I know it's outside the scope of this pull request but we need to clean this up. Would you be able to do that?

jread-usgs:
re: `EVENT_START_DT = '2008-05-20'`
Events won't be multi-month, and this big pile of data seems to dog the line plot. How about `EVENT_START_DT = '2008-06-26'` ?

examples/iowa.py
```diff
@@ -11,10 +11,9 @@
             '05481950', '05416900', '05464500', '05487470']
 
 # Start and end of time series data #
-EVENT_START_DT = '2008-05-20'
+EVENT_START_DT = '2008-06-26'
```
jread-usgs: 👍

floodviz/views.py
```diff
@@ -40,8 +40,8 @@ def timeseries_data():
 def _peakflow_helper():
     # Peak Flow config vars #
     peak_site = app.config['PEAK_SITE']
-    peak_start_date = app.config['PEAK_START_DT']
-    peak_end_date = app.config['PEAK_END_DT']
+    peak_start_date = app.config['EVENT_START_DT']
```
mbucknell: Fine for now. At some point, we should remove the start_date from the peak_utils.req_peak_data function as it is not used.

PR #55, opened 2017-07-12
Refactor code and add function comments

skaymen:
This implements the changes suggested by Mary in pull request 51, namely renaming and commenting the update() function, changing forEach to map, and changing the for loop in click() to data.filter.

### DIFF
--- a/floodviz/static/js/hydrograph.js
+++ b/floodviz/static/js/hydrograph.js
@@ -58,15 +58,16 @@ document.addEventListener("DOMContentLoaded", function (event) {
 				.attr("transform",
 					"translate(" + margin.left + "," + margin.top + ")");
 
-			data.map(function(d) {
-				d.value = Number(d.value);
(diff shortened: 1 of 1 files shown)


floodviz/static/js/hydrograph.js
```diff
@@ -50,7 +58,7 @@ document.addEventListener("DOMContentLoaded", function (event) {
 				.attr("transform",
 					"translate(" + margin.left + "," + margin.top + ")");
 
-			data.forEach(function (d) {
+			data.map(function(d) {
```
mbucknell: So the map function should return the mapped value. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map/
So this should look something like this
```
graphData = data.map(function(d) { 
     return Number(d.value);
}
```
This way you don't change data the parameter passed in. You create a new array for the graph.

PR #59, opened 2017-07-13
Link map to timeseries FV-15

jameeters:
This also includes FV-47: displaying a subset of sites on the hydrograph by default.


floodviz/static/js/hydrograph.js
```diff
@@ -48,35 +56,55 @@
...
-			*
-		*/
-		var update = function(data) {
+
+		/**
+		 * Filters a set of data based on the ids listed in display_ids
+		 * @param data The dataset to be filtered
+		 * @returns {Array} The entries of the original `data` whose `key` values are elements of display_ids.
+		 */
+		var subset_data = function (data) {
```
mbucknell: Consider using the array filter function https://www.w3schools.com/jsref/jsref_filter.asp.

floodviz/static/js/hydrograph.js
```diff
@@ -89,13 +117,12 @@
 				.key(function (d) {
 					return d.key;
 				})
-				.entries(graph_data);
+				.entries(data);
 			// Loop through each symbol / key
 			dataNest.forEach(function (d) {
-				var map_site = document.getElementById('map' + d.key);
-				map_site.classList.add('accent');
+				FV.map_figure.addaccent(d.key);
```
mbucknell: I think Garritt is taking care of this but should not call FV.map_figure. Or alternatively pass in the list of initial gages to the map so that they are accented.

floodviz/static/js/hydrograph.js
```diff
@@ -105,79 +132,121 @@
...
+
 		/**
 		 * Initalize the Hydrograph
 		 */
-		self.init = function() {
+		self.init = function () {
 			var data_path = options.data_path;
 			// Get the data
 			d3.json(data_path, function (error, data) {
+				if (error) { console.error(error); }
```
mbucknell: Is data an empty array in the case of an error? If not this code will through an error when update(data) is called because that is expecting an error. Consider assigning self.full_data = [] and using an else clause.

floodviz/static/js/hydrograph.js
```diff
@@ -105,79 +132,121 @@
...
+		 * Handle all mouse over movements for calling element.
 		 */
-		self.mouseover = function(d) {
-			d3.select(d.data.name).classed("gage--hover", true);
+		self.mouseover = function (d) {
+			self.activate_line(d.data.key);
 			focus.attr("transform", "translate(" + x(d.data.time_mili) + "," + y(d.data.value) + ")");
 			focus.select("text").html(d.data.key + ": " + d.data.value + " cfs " + " " + d.data.time + " " + d.data.timezone);
 			// Interative linking with map
 			FV.map_figure.mousemove(d.data.name, d.data.key);
```
mbucknell: Again, I assume Garrit is handling not using FV.map_figure here.

floodviz/static/js/hydrograph.js
```diff
@@ -105,79 +132,121 @@
...
 			// Interative linking with map
 			FV.map_figure.mouseout();
 		};
-    
+
+		/**
+		 * Handle all click events for calling element.
+		 */
+		self.click = function (d) {
+			FV.map_figure.removeaccent(d.data.key);
```
mbucknell: Ditto.

floodviz/static/js/hydrograph.js
```diff
@@ -105,79 +132,121 @@
...
+		 * @param new_display_ids The new set of gages to be displayed.
 		 */
-		self.click = function(key, data) {
-			d3.select("svg").remove();
-			var new_data = data.filter(function(d) {
-				return d.key !== key;
+		self.change_lines = function (new_display_ids) {
+			display_ids = new_display_ids;
+			var new_data = [];
+			self.full_data.forEach(function (d) {
```
mbucknell: Consider using the array filter function.

floodviz/static/js/map.js
```diff
@@ -1,25 +1,25 @@
...
-		*
-		* mapmodule is a module for creating maps using d3. Pass it a javascript object 
-		* specifying config options for the map. Call init() to create the map. Other pulic fuctions
-		* handle user events and link to other modules. 
-		* 
-	*/
+	 * @param {Object} options - holds options for the configuration of the map.
+	 * All keys are not optional.
+	 * Keys include:
+	 *    'height' v(int) - height of the map
```
mbucknell: This is fine for now. We try to follow jsdoc standards for javascript function comments, http://usejsdoc.org/. Generally, I use @prop indented to describe an object's properties.

floodviz/static/js/map.js
```diff
@@ -61,10 +63,10 @@
...
 					return "translate(" + projection(d.geometry.coordinates) + ")";
 				})
-				.attr("id", function(d) {
-					if (classname === 'gage-point') {
-						return 'map' + d.properties.id;
-					} else { return '';}
+				.attr("id", function (d) {
+					if (property_for_id && d.properties[property_for_id]) {
+						return 'map' + d.properties[property_for_id];
+					}
```
mbucknell: Should be an else clause that returns on an empty string. Also consider using ternary operator, see https://www.w3schools.com/js/js_comparisons.asp rather than an if-else clause.

floodviz/static/js/map.js
```diff
@@ -100,45 +99,66 @@
...
-			var sites = add_circles(options.site_data, "gage-point", 3);
+			var sites = add_circles(options.site_data, "gage-point", 3, 'id');
 			sites.selectAll("circle")
-				.on('mousemove', function(d) {return self.mousemove(d.properties.name, d.properties.id)})
-				.on("mouseout", function() {return self.mouseout()});
+				.on('mousemove', function (d) {return self.mousemove(d.properties.name, d.properties.id)})
+				.on("mouseout", function (d) {return self.mouseout(d.properties.id)})
+				.on('click', function (d) { return self.click(d.properties.id)});
 			// Debug points
 			if (FV.mapinfo.debug) {
```
mbucknell: I wonder if we should just but debug on the FV directly. I think I would like to see config variables in general in there own property. Something like FV.CONFIG = {debug: {{config.DEBUG}}, etc. It's really not mapinfo specific and we may want to add debug features to other parts of the application. Something to consider.

floodviz/static/js/map.js
```diff
@@ -100,45 +99,66 @@
...
 			}
 		};
+
 		/**
-		 * Handle all mouse over movements for calling element. 
+		 * Handle all mouse over movements for calling element.
 		 */
 		self.mousemove = function (sitename, sitekey) {
-			var gage_point_cords = document.getElementById('map'+sitekey).getBoundingClientRect();
+			FV.hydro_figure.activate_line(sitekey);
```
mbucknell: You should talk with Garritt. We want to avoid using global variables in this module patterns. You are tying all potential map figures to this specific hydro_figure. Also this function does not necessarily need to be applied to mousemove. It affects a change on the map visualization and the function name should reflect that.

floodviz/static/js/map.js
```diff
@@ -100,45 +99,66 @@
...
 				.html((sitename));
-			// Link interactions with hydrograph here
-
 		};
 		/**
-		 * Handle all mouse out movements for calling element. 
+		 * Handle all mouse out movements for calling element.
 		 */
-		self.mouseout = function () {
+		self.mouseout = function (sitekey) {
```
mbucknell: Same comments as above.

floodviz/static/js/map.js
```diff
@@ -100,45 +99,66 @@
...
-		self.removeaccent = function(sitekey) {
-			var point = document.getElementById('map' + sitekey);
-			point.classList.remove('accent');
-			// Link interactions with hydrograph here
+		self.removeaccent = function (sitekey) {
+			d3.select('#map' + sitekey).attr('class', 'gage-point');
+		};
+
+
+		self.click = function (sitekey) {
```
mbucknell: Same comments as above.

floodviz/static/js/map.js
jameeters: I added an else clause, but I don't want to use a ternary operator because I find them hard to read.

floodviz/static/js/hydrograph.js
jameeters: I didn't really mean to make the code fault-tolerant, I just wanted it to log an error if it happens.

floodviz/static/js/map.js
mbucknell: I really think it depends on the usage. In this case I think the if else is fine. But I have cases where there are a series of variables that need to have values. Rather than have a series of if else clauses, the code looks a lot cleaner if you use the ternary operator. As with most things, if you get use to it, it looks more natural.

PR #60, opened 2017-07-17
Changes to build and serving with nginx

skaymen:
The application is now frozen during the build process. Nginx is used to serve the static pages. Additionally, a config file can now be passed in during the build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ RUN node_modules/bower/bin/bower install --allow-root
 
 RUN nosetests --all-modules --exe
 
-RUN python3 run.py --config=instance_config.py --freeze --norun
+RUN python3 run.py --config=examples/iowa.py --freeze --norun
 
 EXPOSE 80
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,8 @@ WORKDIR /app
 RUN cp DOIRootCA2.cer /etc/ssl/certs/DOIRootCA2.pem \
 && cat DOIRootCA2.cer >> /etc/ssl/certs/ca-certificates.crt
 
+RUN cp nginx.conf /etc/nginx
+
 RUN rm -f /etc/apt/sources.list.d/chris-lea-node_js-jessie.list \
 && curl --cacert DOIRootCA2.cer https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -28,7 +30,9 @@ RUN node_modules/bower/bin/bower install --allow-root
 
 RUN nosetests --all-modules --exe
 
-EXPOSE 5050
+RUN python3 run.py --config=instance_config.py --freeze --norun
```
jameeters: I think the config file specified should be `examples/iowa.py` so we don't have to have a new file.

run.py
```diff
@@ -28,7 +37,8 @@
         print("freezing")
         freezer.freeze()
         print("frozen")
-        freezer.serve(host=host, port=port, threaded=True)
+        if not no_run:
```
jameeters: If you  change line 18 to `run = !args.norun` then this could change to `if  run:`, which I think it much easier to understand. But this is alright and I'm not determined to change it.

README.md
```diff
@@ -7,14 +7,14 @@ the frozen files.
 
 **To build the image:**
 1. Navigate to your project directory.
-1. Put your configuration parameters in `instance/config.py`
+1. Put your configuration parameters in 'instance_config.py' and copy into this directory.
```
skaymen: Removed.

Dockerfile
skaymen: Changed.

run.py
skaymen: Fixed.

mbucknell:
The Dockerfile build would still require the config file to be in the directory that is copied to the container. Can we change this behavior so that the config file can be copied from anywhere in the builder's file system into the container?

nginx.conf
```diff
@@ -0,0 +1,90 @@
...
+    #keepalive_timeout  0;
+    keepalive_timeout  65;
+    # timeout during which a keep-alive client connection will stay open.
+
+    #gzip  on;
+    # tells the server to use on-the-fly gzip compression.
+
+    server {
+        location / {
+            root /app/floodviz/build/;
```
ayan-usgs: I think it'd be less confusing for future developers to know where the `build` directory is coming from. Folks who look at the requirements and see Frozen-Flask will realize where the build directory is coming from, but it does require some sleuthing and familiarity with the project; it new developer working on the NGINX config sometime to figure all that out.

skaymen: Fixed.

PR #61, opened 2017-07-17
Merge interactions fv 15 and 16

jameeters:
Combined changes for FV-15 (link hydrograph to map) and FV-16 (link map to hydrograph). 
Also includes FV-48 (refactor `hydrograph.js` and `map.js` to use the module pattern).


floodviz/static/js/map.js
```diff
@@ -100,45 +110,68 @@
 			add_paths(options.rivers_data, "river");
 			add_circles(options.ref_data, "ref-point", 2);
 			// Add sites and bind events for site hovers
-			var sites = add_circles(options.site_data, "gage-point", 3);
+			var sites = add_circles(options.site_data, "gage-point", 3, 'id');
 			sites.selectAll("circle")
-				.on('mousemove', function(d) {return self.mousemove(d.properties.name, d.properties.id)})
-				.on("mouseout", function() {return self.mouseout()});
+				.on('mousemove', function (d) {
```
skaymen: It may be better to use mouseover instead of mousemove here.

floodviz/static/js/floodviz_onready.js
```diff
@@ -1,28 +1,48 @@
...
+		'width': FV.hydrograph_dimensions.width,
+		'div_id': '#hydrograph'
+	};
+
+	var map_frame = FV.mapmodule(map_options);
+	var hydro_frame = FV.hydromodule(hydro_options);
+
+
+	// Use frames to link interactions
+	hydro_options['site_tooltip_show'] = map_frame.site_tooltip_show;
```
mbucknell: I know this works so don't feel like you need to change this. I wonder if the code would be more readable/understandable if you did something like this:
```
var this_site_tooltip_remove;
var hydro_options = {
    site_tooltip_remove : this_site_tooltip_remove;
};
```
Then create your hydro_frame and map_frame and then assign this_site_tooltip_remove = ma_frame.site_tooltip_remove;

mbucknell: The other option we might consider is passing in the functions to the init function. This might be clearer

ede0m: I'm not sure if you have tried tried something like the first option you suggested before, but I cannot get it to work at the moment.

mbucknell: With further thought, I like my second option better where we pass the functions into the init. Probably still pass it in as an Object with properties rather than individual parameters.

PR #62, opened 2017-07-18
Travis setup fv 51

jameeters:


.travis.yml
```diff
@@ -0,0 +1,9 @@
+dist: trusty
+language: python
+python: "3.5"
```
mbucknell: Can we use python: 3.6?

jameeters: We're now testing against 3.5 and 3.6.

mbucknell:
Testing both is probably overkill for our purposes. But fine for now.

PR #63, opened 2017-07-18
Parse utils functions and tests - FV_40

ede0m:


floodviz/utils.py
```diff
@@ -0,0 +1,90 @@
...
+"""
+    Takes a valid url and a dictionary of expected column header strings as keys (values do not matter).
+    parse_rdb then requests data from the service at url, and extracts data matching any header values.
+     
+    Data is returned as a list of dict objects, where each object represents a data point. 
+
+"""
+
+
+def parse_rdb(url, header_values):
```
jameeters: `header_values` should be passed in as a List if the values don't matter. You can always transform it into a dict of blank entries with `d = dict(zip(lst, ['' for _ in lst]))`, but here it's only being used as a list.

floodviz/utils.py
```diff
@@ -0,0 +1,90 @@
...
+        keep_headers = {}
+        found_headers = False
+        for line in content:
+            if line.startswith('#'):
+                continue
+            # should run once. Used to map column headers to line split indices.
+            if not found_headers:
+                raw_headers = line.split('\t')
+                for idx, key in enumerate(raw_headers):
+                    if header_values.get(key) is not None:
```
jameeters: This will fail if the value is `None`. A better solution is `if key in header_values.keys()`.

floodviz/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    try:
+        r = requests.get(url)
+    except requests.exceptions.RequestException as e:
+        print('- Bad URL -')
+    else:
+        if r.status_code is 200:
+            content = r.text.splitlines()
+    if content:
+        all_data = []
+        keep_headers = {}
```
jameeters: I don't see why you avoid using `header_values` to store the positions of the headers. It's fine to use a new dictionary, but this way there's no reason for `header_values` to be a dictionary because only its keys are ever used.

floodviz/utils.py
```diff
@@ -0,0 +1,90 @@
...
+            # should run once. Used to map column headers to line split indices.
+            if not found_headers:
+                raw_headers = line.split('\t')
+                for idx, key in enumerate(raw_headers):
+                    if header_values.get(key) is not None:
+                        keep_headers[key] = idx
+                found_headers = True
+                continue
+            else:
+                if line.startswith('USGS'):
```
jameeters: There are gages that aren't run by the USGS, so we should make sure we don't need to be able to use those.

floodviz/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    if params.get('start_date_query'):
+        url += params['start_date_query'] + params['start_date']
+    if params.get('end_date_query'):
+        url += params['end_date_query'] + params['end_date']
+    # one or the other. Not both. 'site_status' and 'agency' keys should not be in the same params dict.
+    if params.get('site_status'):
+        url += params['site_status']
+    elif params.get('agency'):
+        url += params['agency']
+    return url
```
jameeters: I don't think this function is likely to make anything easier. There's very few common components between the URLs for all our different requests, so I think it'll be easier to handle URL construction on a case-by-case basis. Constructing the URLs this way would make more sense if we were making a bunch of calls with very similar parameters.

floodviz/utils.py
```diff
@@ -0,0 +1,90 @@
...
+def parse_rdb(url, header_values):
+    content = None
+    try:
+        r = requests.get(url)
+    except requests.exceptions.RequestException as e:
+        print('- Bad URL -')
+    else:
+        if r.status_code is 200:
+            content = r.text.splitlines()
+    if content:
```
jameeters: Isn't this (line 21) essentially the same as `if r.status_code is 200:`, since `content` is never set if the status is not 200?

floodviz/utils.py
mbucknell: I was thinking of a much more generic parse function: Something like
parse_rdb(url, params) where url is the non query string part of the request and params are the query parameters without including the format for rdb (that would be added) by this function.  I would pass back all columns. Let the caller of the function decide what to do with the dictionary.

ede0m: I made changes to parse_rdb based on these comments. Let me know what you think when you get a chance. I plan to make the data_column extractor in the near future.

floodviz/utils.py
```diff
@@ -0,0 +1,82 @@
...
+                    # extraction procedure
+                    line = line.split('\t')
+                    data_point = {}
+                    # sanity check that headers has been filled
+                    if headers is not None:
+                        for idx, data in enumerate(line):
+                            data_point[headers[idx]] = data
+                        all_data.append(data_point)
+                # Should run before above if block
+                elif headers is None:
```
jameeters: I think you should be more explicit about deciding which line will be used as the headers. I think this should use the first non-blank line after the block of comments at the beginning. I think it would be better to skip all lines that start with `#` or are blank, then use the next line as the header, and then move on to the parsing of actual content. I don't like the `if headers if not None` before the line parsing. 

This is also entirely personal preference, so feel free to ignore it.

floodviz/utils.py
```diff
@@ -0,0 +1,82 @@
...
+                if line.startswith('#'):
+                    continue
+                if line.startswith('USGS'):
+                    # extraction procedure
+                    line = line.split('\t')
+                    data_point = {}
+                    # sanity check that headers has been filled
+                    if headers is not None:
+                        for idx, data in enumerate(line):
+                            data_point[headers[idx]] = data
```
jameeters: This seems equivalent to:
`data_point = dict(zip(headers, line))`

It looks like this (either as-is or using `zip`) will fail if any field is missing from the line. I'm not sure if we have to take that into account, but I'd like to be sure that the parser will either always return correct data or fail* rather than returning bad data.

*Correct meaning that it can assign the right value to the right header even if a value is missing from the line, and fail meaning either throw an error or return None.

floodviz/tests/test_utils.py
```diff
@@ -0,0 +1,416 @@
...
+            'site_query': '&sites=',
+            'start_date': self.peak_dv_date,
+            'start_date_query': '&startDT=',
+            'end_date': self.peak_dv_date,
+            'end_date_query': '&endDT=',
+            'format': '?format=rdb',
+            'site_status': '&siteStatus=all',
+            'extra': 'dv'
+        }
+        self.valid_return_peak = [
```
jameeters: It seems like it would be easier if these test cases weren't so long, but it's probably not worth the work to shorten them.

floodviz/utils.py
```diff
@@ -0,0 +1,82 @@
...
+                    'agency' - service specific string for url construction. Cannot be used with 'site_status' key
+
+        RETURNS:
+            url - url to request for data
+    """
+    all_data = []
+    sites = None
+    url = endpoint
+    data_format = '?format=rdb'
+    if params.get('sites') is not None:
```
mbucknell: My intention was to have the caller know what query parameters to use so that this is truly and general purpose utility. The params should be passed directly to the requests.get after adding the format query parameter so that the output is rdb. We shouldn't be doing a check here to see if the query parameters are valid or whether they make sense.

floodviz/utils.py
```diff
@@ -0,0 +1,82 @@
...
+    elif params.get('agency'):
+        url += params['agency']
+    # request url
+    try:
+        r = requests.get(url)
+    except requests.exceptions.RequestException as e:
+        print(e)
+        print('- bad webservice url -')
+    else:
+        if r.status_code is 200:
```
ede0m: I like this style too, but was requested to try to limit functions to one return statement. That was my reasoning for inverting the condition.

mbucknell: We try to follow a best practice here that functions should only have a single return statement.

floodviz/utils.py
```diff
@@ -0,0 +1,82 @@
...
+            content = r.text.splitlines()
+            headers = None
+            for line in content:
+                # Skip commented lines
+                if line.startswith('#'):
+                    continue
+                if line.startswith('USGS'):
```
ede0m: Good point here.

floodviz/utils.py
ede0m: Tests confirm the validity of the data format. If data is missing from a column for any row, the data value for that datapoint key is ''.

floodviz/utils.py
ede0m: @mbucknell, I think I understand what you are getting at here. Similar to 'passing parameters to url' at this [link?](http://docs.python-requests.org/en/master/user/quickstart/)

mbucknell: Yes.

floodviz/utils.py
```diff
@@ -0,0 +1,54 @@
...
+            endpoint - String specifying the web service url
+            params - dictionary holding url parameters
+                * keys for params should be the value you variable name for the query parameter
+                  you want to specify. Values for these keys should be the query parameter value
+                  you wish to use.
+
+                  i.e {'site_no' : '056777781'} -> '&site_no=056777781'
+
+        RETURNS:
+            url - url to request for data
```
mbucknell: It is not returning  url. This needs to be corrected.

floodviz/utils.py
```diff
@@ -0,0 +1,54 @@
...
+            headers = None
+            for line in content:
+                # Skip commented lines
+                if line.startswith('#'):
+                    continue
+                if headers is None:
+                    # header procedure
+                    line = line.split('\t')
+                    headers = line
+                    # Skip weird '5s      15s     20d     14n     10s\n'\' which appears on next line after headers
```
mbucknell: Please no editorial comments

floodviz/utils.py
```diff
@@ -0,0 +1,54 @@
...
+    except requests.exceptions.RequestException as e:
+        print(e)
+        print('- bad webservice url -')
+    else:
+        if r.status_code is 200:
+            content = iter(r.text.splitlines())
+            headers = None
+            for line in content:
+                # Skip commented lines
+                if line.startswith('#'):
```
mbucknell: This statement will skip any line that begins with a # sign. After the beginning comments and the header lines, the rest of the lines are data and may begin with a # sign. This needs to be written so that only the initial lines with a # sign are skipped. Considering using a while loop to remove the initial comment lines and then proceed.

floodviz/utils.py
mbucknell: Also recommend using data_point = dict(zip(headers, line)) rather than this for loop

floodviz/tests/test_utils.py
```diff
@@ -0,0 +1,407 @@
+import unittest
+import requests_mock
+from floodviz.utils import parse_rdb
+
+
+class TestPeakUtilsRDBparse(unittest.TestCase):
```
mbucknell: The convention for naming tests classes is TestCamelCasedMethodName so this should be TestParseRdb.

The tests need to be updated. There are properties on self that are not used in the tests. I would recommend simplifying your self.response to be a few lines of comments, a few columns, and then three or four data lines. It does not need to mimic a real response. You will also want to have a test where the  response includes no data lines.

floodviz/utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def parse_rdb(endpoint, params):
+    """
+        Takes a web service endpoint, a dictionary of key/value pairs for the query parameters
+        and returns all parsed data from the webservice.
+
+        ARGS:
+            endpoint - String specifying the web service url
+            params - dictionary holding url parameters
```
mbucknell: Nit: Should say dictionary holding query parameters.

floodviz/utils.py
ede0m: what about 
`if line.startswith("#') and headers is None: `

I believe this accomplishes the same thing as a while loop skipping until headers and found.

floodviz/tests/test_utils.py
```diff
@@ -0,0 +1,210 @@
+import unittest
+import requests_mock
+from floodviz.utils import parse_rdb
+
+
+class TestRdbParsePeak(unittest.TestCase):
```
mbucknell: Please look at the previous comments for tests. Please follow the convention of the test class being TestRdbParse (TestCamelCaseFunctionBeingTested). You should only need a single TestCase class. Also please include a test for a response where no data lines are returned.

As I said before, it is not necessary to exactly mimic a call that you are planning to make nor mimic a response that you will get. The test will be easier to read, if you mock a response that has a few lines of comments, just a few columns and three or four rows of data.

floodviz/tests/test_utils.py
```diff
@@ -0,0 +1,210 @@
...
+            m.get(self.mock_url_peak, text=self.response_peak)
+            self.assertEqual(parse_rdb(self.endpoint_mock, self.test_dict_peak), self.valid_return_peak)
+
+    def test_valid_data_dv(self):
+        with requests_mock.Mocker() as m:
+            m.get(self.mock_url_dv, text=self.response_dv)
+            self.assertEqual(parse_rdb(self.endpoint_mock, self.test_dict_dv), self.valid_return_dv)
+
+
+class TestRdbParseMap(unittest.TestCase):
```
mbucknell: Please remove this. Not sure what is tested here that wasn't tested as part of the first test case.

floodviz/utils.py
```diff
@@ -0,0 +1,52 @@
...
+            endpoint - String specifying the web service url
+            params - dictionary holding query parameters
+                * keys for params should be the value you variable name for the query parameter
+                  you want to specify. Values for these keys should be the query parameter value
+                  you wish to use.
+
+                  i.e {'site_no' : '056777781'} -> '&site_no=056777781'
+
+        RETURNS:
+            list of data point dicts extracted from webservice.
```
mbucknell: You should say that it returns None if the webservice call fails.

floodviz/utils.py
```diff
@@ -0,0 +1,52 @@
...
+                    # extraction procedure
+                    line = line.split('\t')
+                    # sanity check that headers has been filled
+                    if headers is not None:
+                        data_point = zip(headers, line)
+                        all_data.append(dict(data_point))
+        else:
+            all_data = None
+
+    print(all_data)
```
mbucknell: Please remove this print statement.

floodviz/tests/test_utils.py
ede0m: I was assuming we wanted tests for every service this would be used for, as the responses are different between them. But if this isn't the case, I will remove them.

floodviz/tests/test_utils.py
ede0m: With one test case class, should I include the tests for the webservice that the map_utils currently uses (TestRdbParseMap)? Or should I remove the test for daily value?

I did already remove a good portion of the response, but I will remove more.

floodviz/tests/test_utils.py
mbucknell: The unit test should only test the particular unit, not how it will be used in the application. The unit test does not need to reflect a "real" example but should mimic the important pieces. In this case all we need to do is test whether it parses a file in rdb format, not necessarily an actual response.

floodviz/tests/test_utils.py
mbucknell: See my comment from below. You don't need to test how the rest of the application will potentially be using the function. This is a general purpose function and should be able to handle any web service that returns an rdb format.

PR #70, opened 2017-07-19
pass linked interactions to init -- FV_48

ede0m:

PR #71, opened 2017-07-19
Map box selection fv 45

jameeters:


floodviz/static/js/map.js
```diff
@@ -1,187 +1,321 @@
...
-				.attr("class", classname)
-				.each(function (d) {
-					if (property_for_id && d.properties[property_for_id]) {
-						if (FV.hydrograph_display_ids.indexOf(d.properties.id) !== -1) {
-							self.site_add_accent(d.properties.id);
+		FV.mapmodule = function (options) {
+
+			var self = {};
+
+			var state = {};
```
ede0m: Comment for state? I'm guessing this is specific to the select drag box?

floodviz/static/js/map.js
```diff
@@ -1,187 +1,321 @@
...
+						self.site_tooltip_show(d.properties.name, d.properties.id);
+						options.hover_in(d.properties.id);
+					})
+					.on("mouseout", function (d) {
+						self.site_tooltip_remove(d.properties.id);
+						options.hover_out(d.properties.id);
+					})
+					.on('click', function (d) { self.click(d.properties.id)});
+
+				state.gages = [];
```
ede0m: Comment for code below. Saving all coordinates of gauges for later comparison.

floodviz/static/js/map.js
```diff
@@ -1,187 +1,321 @@
...
+
+				state.box = {
+					x: p[0],
+					y: p[1],
+					height: 0,
+					width: 0
+				};
+			};
+
+			var select_box_drag = function (p) {
```
ede0m: Maybe move next to other private functions?

floodviz/static/js/map.js
```diff
@@ -1,187 +1,321 @@
...
+					state.box = {
+						x: d.x,
+						y: d.y,
+						height: d.height,
+						width: d.width
+					};
+				}
+			};
+
+			var select_box_end = function () {
```
ede0m: Maybe move by other private functions. + Comment for this func.

floodviz/static/js/map.js
```diff
@@ -1,187 +1,321 @@
...
+				// Check if the box has a reasonable area to make sure it isn't a click by mistake
+				var area = state.box.width * state.box.height;
+				if (area >= 100) {
```
ede0m: Should we define this 100 somewhere as a var? May be unnecessary though.

floodviz/static/js/map.js
```diff
@@ -1,187 +1,321 @@
...
+						x: state.box.x,
+						y: state.box.y
+					};
+					var SE = {
+						x: NW.x + state.box.width,
+						y: NW.y + state.box.height
+					};
+					var selected = [];
+					FV.hydrograph_display_ids.forEach(function (key) {
+						self.site_remove_accent(key);
```
ede0m: Does this remove all accents before adding later on? Could we just have a condition from 308-310 where we check if the site already has accent class?

floodviz/static/js/map.js
```diff
@@ -1,187 +1,321 @@
...
+				var b = path.bounds(options.bounds);
+				var s = options.scale / Math.max((b[1][0] - b[0][0]) / options.width, (b[1][1] - b[0][1]) / options.height);
+				var t = [(options.width - s * (b[1][0] + b[0][0])) / 2, (options.height - s * (b[1][1] + b[0][1])) / 2];
+				// Update the projection
+				projection.scale(s).translate(t);
+				// Add layers
+				var bg =add_paths(options.bg_data, "background");
+				bg.call(drag);
+				var rivers = add_paths(options.rivers_data, "river");
+				rivers.call(drag);
```
ede0m: Not entirely sure, but do we need to call drag on rivers too?

jameeters: It lets the user click and drag on a river, which will be nice if the rivers are thicker than they are in the Iowa example.

floodviz/static/js/map.js
jameeters: This is mostly there to make sure none of the sites that we remove keep their accents. I could do it with a conditional later on, but I'd still have to go through all the displayed sites at some point and remove the accent for any not in the selection, so I think it boils down to the same thing.

floodviz/static/js/map.js
jameeters: I don't think this is something that we're going to need to change often, and it's only used once, so I'm going to leave it as-is

floodviz/static/js/map.js
```diff
@@ -1,187 +1,331 @@
...
+			};
+
+			/**
+			 * End the selection box, get rid of the box in state and on the svg,
+			 * update the hydrograph and map if necessary
+			 */
+			var select_box_end = function () {
+				// Check if the box has a reasonable area to make sure it isn't a click by mistake
+				var area = state.box.width * state.box.height;
+				if (area >= 100) {
```
mbucknell: I was looking at https://github.com/d3/d3-drag/blob/master/README.md#drag_clickDistance to see if we can use it it to define a click for a gage vs mouse down-drag for the draw box. Have you looked at this? This might make it easier to apply the drag interaction to the whole svg instead of layer by layer.

floodviz/static/js/map.js
```diff
@@ -1,187 +1,331 @@
...
+					})
+					.on("mouseout", function (d) {
+						self.site_tooltip_remove(d.properties.id);
+						options.hover_out(d.properties.id);
+					})
+					.on('click', function (d) { self.click(d.properties.id)});
+
+				// Save locations of gages in SVG for later use with selection box
+				state.gages = [];
+				options.site_data.features.forEach(function (g) {
```
mbucknell: Consider using the filter method to create state.gages.

floodviz/static/js/map.js
jameeters: I just tried it and couldn't get it to work, but it works as-is when I apply it to the whole svg.

floodviz/static/js/map.js
```diff
@@ -1,193 +1,357 @@
...
+					.attr("d", path)
+					.attr("class", classname);
+				return group;
+			};
+
+			/**
+			 * Add or remove the line corresponding to a gage from the hydrograph
+			 * @param sitekey THe ID of the gage in question
+			 */
+			var toggle_hydrograph_display = function (sitekey) {
```
mbucknell: It seems like we are targeting the hydrograph specifically again. We may need to revisit the linked_interactions so that the linked_nteractions are abstracted better. This is okay for now.

jameeters: This is the same as the `click` function from #70 (FV-48). I made it a private function and renamed it, but I didn't (intentionally) make any other changes.

mbucknell: I noted on that pull request that there still is something not quite right with the way we are abstracting this. I may try give it some thought when I have some time. Feel free to see what you can come up with too if you want.

PR #77, opened 2017-07-21
Reset hydrograph on double click fv 58

jameeters:


floodviz/static/js/hydrograph.js
```diff
@@ -63,6 +65,19 @@
 		};
 
 		/**
+		 * Show only the default set of lines on the hydrograph.
+		 */
+		var reset_hydrograph = function () {
+			FV.hydrograph_display_ids.forEach(function (id) {
```
mbucknell: I should have caught this before. Can we pass this in through the options? I think you won't need default_display_ids, it could just be options.default-display_ids.

We talked about this and I will make a separate ticket.

PR #78, opened 2017-07-25
Deemphasis interaction fv 15 and fv 17

jameeters:


floodviz/static/js/hydrograph.js
```diff
@@ -62,6 +62,15 @@
 			return toKeep;
 		};
 
+		var make_lines_bland = function (key){
```
ede0m: Just a little comment.

floodviz/static/js/map.js
```diff
@@ -236,6 +232,34 @@
...
+					if (g.accent) {
+						style += '-accent'
+					}
+					if (g.bland) {
+						style += '-bland';
+					}
+					else {
+						g.bland = true;
+					}
+					d3.select('#map' + key).attr('class', style);
```
ede0m: I like the style class appending here! But the function name doesn't necessarily represent the actual functionality well if logic has the possibility to accent and not make bland here. 

`if (g.accent) {style += '-accent'}` is a part of make_sites_normal. Why use it here as well? Also do we want to be setting g.bland to true if it was already false? Is this preemptive?

jameeters: What this does is make all but one of the sites bland. When you moseover a site `123`, `state.gages[123].bland` is set to `false` so that it will be exempted here. I set it back to `true` here to make sure it can be set to bland next time. I did it this way because I wanted to make it easy to have multiple sites exempted if we want to add that in the future, but I'd be open to changing it. It would probably be easier for now to just pass a single site ID that we want to exempt.

There are two 'bland' styles `gage-point-bland` and `gage-point-accent-bland` so that the sites displayed on the hydrograph are still distinguished from the sites that aren't. This won't change whether or not a gage is accented, it just checks which bland style it should apply to each.

floodviz/static/js/map.js
```diff
@@ -223,14 +216,17 @@
 					};
 					var selected = FV.hydrograph_display_ids;
 
-					state.gages.forEach(function (g) {
+					const keys = Object.keys(state.gages);
```
mbucknell: Like the use of const!

PR #79, opened 2017-07-25
Thumbnail fv_34 with JSDOM

ede0m:
There is still plenty to do with this ticket. Namely, lots of refactoring once we start using reference.json. Right now we are using almost a direct copy of hydromodule, with a few minor differences to make this work post build. Ideally we would like to not have a duplicate file like this.

New dependencies:

    npm install jsdom --save
    npm install svg2png --save

Usage (with node version 7.5.0): 

    node floodviz/posterior_build/thumbnail.js -figureType

_Valid figure types include:_ -hydro, -map


floodviz/posterior_build/hydro_thumbnail.js
```diff
@@ -0,0 +1,248 @@
+
```
skaymen: Could use comments indicating this is same as hydrograph code and what differences are.

floodviz/posterior_build/map_thumbnail.js
```diff
@@ -0,0 +1,363 @@
+
```
skaymen: Could use similar comment to hydrograph.

floodviz/posterior_build/thumbnail.js
```diff
@@ -0,0 +1,92 @@
+var fs = require('fs');
```
skaymen: A comment here would help explain what these variables will be used for.

floodviz/posterior_build/thumbnail.js
```diff
@@ -0,0 +1,92 @@
...
+	// load local assets into window environment (!)
+	[
+		'./floodviz/static/bower_components/d3/d3.js',
+		'./floodviz/static/bower_components/jquery/dist/jquery.min.js',
+		'./floodviz/static/bower_components/proj4/dist/proj4.js',
+		'./floodviz/posterioir_build/hydro_thumbnail.js',
+		'./floodviz/posterioir_build/map_thumbnail.js'
+	],
+
+	function (err, window) {
```
skaymen: Function comments would be helpful here

ede0m:
I verified that this does not work with Node version 6.11.0. If we want to use this or a variant of this implementation, we will have to use node >= 7.5

ede0m:
Was just thinking that adding an optional argument for external css files may be useful. Let me know what you guys think and I can add this if desired.

floodviz/views.py
```diff
@@ -34,6 +34,10 @@ def timeseries_data():
     # Hydrodata data clean and write
     j = hydrograph_utils.req_hydrodata(sites, hydro_start_date, hydro_end_date, url_nwis_prefix)
     timeseries_data = hydrograph_utils.parse_hydrodata(j)
+
+    # Write data to file for post build thumbnail
+    with open('floodviz/static/data/hydrograph_data.json', 'w') as f:
```
mbucknell: This should not be written to the static directory because it is not data used by the web site. For now let's write it to the new floodviz/thumbnail directory.

mbucknell: I'd also like this to happen only when we ask it to be generated. maybe through a config.py variable

floodviz/views.py
```diff
@@ -94,4 +98,8 @@ def _map_helper():
             ]
         }
     })
+
+    with open('floodviz/static/data/map_data.json', 'w') as f:
```
mbucknell: Same comments as above.

floodviz/posterior_build/hydro_thumbnail.js
ede0m: agreed

floodviz/posterior_build/hydro_thumbnail.js
```diff
@@ -0,0 +1,248 @@
...
+ *        @prop 'width' v(int) - width of the graph
+ *        @prop 'data' v(list) - A list of objects representing data points
+ *        @prop 'div_id' v(string) - id for the container for this graph
+ *
+ * hydromodule is a module for creating hydrographs using d3. Pass it a javascript object
+ * specifying config options for the graph. Call init() to create the graph. Linked
+ * interaction functions for other figures should be passed to init in and object.
+ *
+ */
+var hydromodule = function (options) {
```
ede0m: Are you saying change the original hydrograph.js to not be wrapped in a function, and then we wont have to replicate this file? I will definitely try this, but I am still thinking we will have to battle the FV.hydromodule issue with the original file.

floodviz/views.py
ede0m: Good idea.

floodviz/posterior_build/hydro_thumbnail.js
mbucknell: Maybe. The alternate is to not use global namepsace of FV in the original module. I would rather do that the duplicate code.

ede0m: Wondering if we could make another ticket for this and attack it once the reference.json changes are merged.

ede0m: Actually, it looks like they are already merged. A separate ticket for this would still be helpful IMO. There may be a decent amount of refactoring. Something to talk about at sprint planning I guess.

mbucknell: Let's talk about at sprint planning. I'm okay with putting in a separate ticket though.

floodviz/thumbnail/thumbnail.js
```diff
@@ -0,0 +1,90 @@
...
+	"<html><body><div id='hydrograph'></div>" +
+	"<div id='map'></divid>" +
+	"</body></html>",
+
+	// load local assets into window environment
+	[
+		'./floodviz/static/bower_components/d3/d3.js',
+		'./floodviz/static/bower_components/proj4/dist/proj4.js',
+		'./floodviz/thumbnail/hydro_thumbnail.js',
+		'./floodviz/thumbnail/map_thumbnail.js'
```
mbucknell: Please remove the map_thumbnail.js

floodviz/thumbnail/thumbnail.js
```diff
@@ -0,0 +1,90 @@
...
+				'height': 300,
+				'width': 560,
+				'div_id': '#hydrograph',
+				'data': data_hydro,
+				"display_ids": ['05471200', '05476750', '05411850', '05454220',
+				 '05481950', '05416900', '05464500', '05487470']
+				// Refactor Later. I'm assuming this will change with references.json
+			}
+		);
+		convert(hydro_figure,window,'floodviz/static/css/hydrograph.css','floodviz/thumbnail/thumbnail_hydro.png');
```
mbucknell: Should be spaces after ','

floodviz/thumbnail/thumbnail.js
```diff
@@ -0,0 +1,90 @@
...
+	}
+);
+
+// Wrapper around svg2png that injects custom css to inline svg before conversion
+function convert(figure, window, css_path, filename) {
+	figure.init(undefined);
```
mbucknell: It's better style to put this after the variable declarations. You can write this as figure.init().

floodviz/views.py
```diff
@@ -91,4 +96,9 @@ def _map_helper():
             ]
         }
     })
+
+    if thumbnail:
```
mbucknell: Please remove. Let's stick to just the hydrograph.

package.json
```diff
@@ -18,5 +18,9 @@
   "homepage": "https://github.com/USGS-VIZLAB/active-flood-viz#readme",
   "devDependencies": {
     "bower": "1.8.0"
+  },
+  "dependencies": {
+    "jsdom": "11.1.0",
```
mbucknell: These are actually dev dependencies since they are only used in the build process.

PR #80, opened 2017-07-25
Read in config info from reference.json

skaymen:
Read in the config and geographic information contained in reference.json instead of config


floodviz/views.py
```diff
@@ -52,19 +54,19 @@ def _peakflow_helper():
...
-    with open(app.config['RIVERS_FILE'], 'r') as rivers_file:
-        rivers = json.load(rivers_file)
+    bg_data = json.loads(ref.background_data)
+    bg_data = map_utils.filter_background(ref.bbox, bg_data)
 
-    ref_data = app.config['REFERENCE_DATA']
+    rivers = json.loads(ref.river_data)
 
+    for feature in rivers['features']:
+        del(feature["id"])
```
ede0m: I'm probably missing something here, but why do we have to delete data?

skaymen: Left in by mistake. Good catch.

floodviz/reference_parser.py
```diff
@@ -0,0 +1,39 @@
+import json
+
+# open up and reference file and store the relevant data in variables
+with open('floodviz/static/reference/reference.json', 'r') as f:
```
mbucknell: I think this should be made into a callable function so that it's execution is explicit, rather than implicit by merely importing this python module. I think it makes sense for it to return the parsed information in a dictionary. It can then be called from floodviz/__init__.py before importing the views. You will still need to assign the result of the function to a global variable but it would be scoped to the floodviz package. The other option would be to create a python class. If you did this I would still have a function that explictly reads the configuration information from the file so the caller can control when this information is read.

I think it also makes sense to pass in the location of the reference.json as a parameter to the function. This gives us more flexibility done the road.

I think it also makes sense to put the reference.json in the examples directory for now. The build process should include a parameter for this file's location as we do with the instance/config.py.

floodviz/reference_parser.py
```diff
@@ -0,0 +1,39 @@
...
+    end_date = data['endDate']
+    peak_dv_date = data['endDate']
+    peak_site = data['peak']['site']
+
+    # grab the cities to be placed on the map
+    city_data = []
+    for i in range(0, len(data['reference']['features'])):
+        if data['reference']['features'][i]['properties']['reftype'] == 'city':
+            city_data.append(data['reference']['features'][i])
+    city_data = {"type": "FeatureCollection", "features": city_data}
```
mbucknell: Consider assigning the result of the filtering to a different variable then city_data. I believe this makes the code easier to understand.

floodviz/reference_parser.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # grab the rivers to be placed on the map
+    river_data = []
```
mbucknell: Same comment as for city_data.

floodviz/reference_parser.py
```diff
@@ -0,0 +1,39 @@
...
+    for i in range(0, len(data['reference']['features'])):
+        if data['reference']['features'][i]['properties']['reftype'] == 'rivers':
+            river_data.append(data['reference']['features'][i])
+    river_data = json.dumps({"type": "FeatureCollection", "features": river_data})
+
+    # grab the political/state borders to be placed on the map
+    background_data = []
```
mbucknell: Same comment as for city_data

floodviz/static/css/map.css
```diff
@@ -50,6 +50,7 @@
 #map path.river{
 	stroke: steelblue;
 	fill: steelblue;
+	fill-opacity: 0;
```
mbucknell: Why are you setting this to 0? Don't we want fill?

floodviz/reference_parser.py
skaymen: Fixed.

floodviz/reference_parser.py
skaymen: Fixed.

floodviz/reference_parser.py
skaymen: Fixed.

floodviz/reference_parser.py
```diff
@@ -0,0 +1,43 @@
...
+        parsed_data['display_sites'] = data['display_sites']
+        parsed_data['bbox'] = data['bbox']
+        parsed_data['start_date'] = data['startDate']
+        parsed_data['end_date'] = data['endDate']
+        parsed_data['peak_dv_date'] = data['endDate']
+        parsed_data['peak_site'] = data['peak']['site']
+
+        # grab the cities to be placed on the map
+        city_geojson_data = []
+        for i in range(0, len(data['reference']['features'])):
```
mbucknell: Consider using a list comprehension with an if (https://docs.python.org/3/tutorial/datastructures.html?highlight=comprehension#list-comprehensions).I made this comment before but forgot to hit Add. Sorry about that. Same comment applies to rivers and background.

skaymen: Fixed.

PR #81, opened 2017-07-26
updating README w/ docker commands

jread-usgs:
Suggestions for README changes


jread-usgs:
Also, curious why you don't need to run `docker run --rm -dp 80:80 -t flood:latest` as sudo, since you are using 80:80.

PR #84, opened 2017-07-26
Implement changes suggested by jread

skaymen:


ede0m:
Don't forget the snoo! 
snoo

PR #85, opened 2017-07-26
small config updates

jread-usgs:
To center on the flood and feature a better gage


mbucknell:
I think this will be obsolete soon (I think we are close to merging the code using the reference.json) and the parameters changed are all in reference.json.

jread-usgs:
no prob. I can PR a change in reference.json when that work is merged

PR #91, opened 2017-07-28
The timeseries view should be mapped to timeseries.json

mbucknell:


ede0m:
Did this get rid of that MIMETYPE warning we were getting during the build?

mbucknell:
Yes. I was also hoping it would take care of the issue I'm having pushing the code out to CHS, but it did not fix that.

PR #93, opened 2017-07-31
Refac thumbnail fv 73

ede0m:
Mostly involved adding reading in default_display_ids from examples into thumbnail.js

Can run and test the node script with the same command used in [this pull request](https://github.com/USGS-VIZLAB/active-flood-viz/pull/79)


floodviz/thumbnail/thumbnail.js
```diff
@@ -82,7 +83,7 @@ function convert(figure, window, css_path, filename) {
 		svg_string = inject_style(style_default, null, svg, window);
 	}
 	// Takes care of canvas conversion and encodes base64
-	svg2png(svg_string)
+	svg2png(svg_string, {height: 300, width: 560})
```
skaymen: The numbers for height and width are used twice in this file-- maybe we could make them vars?

ede0m: There was refactoring done to height and width within the figure module while I was working on implementing the thumbnail script. When jumping on this current ticket, height and width had been removed from hydrograph.js during svg initialization (lines 116-122). This required me to specify height in width in svg2png because it could not be interpreted from the svg itself.

ede0m: I did however, take your suggestion and declare vars at the top for height and width instead of using the actual integers.

floodviz/thumbnail/thumbnail.js
```diff
@@ -15,6 +15,7 @@ var jsdom = require('jsdom/lib/old-api.js');
 var svg2png = require('svg2png');
 // Data imports
 var data_hydro = require('../thumbnail/hydrograph_data.json');
+var reference = require('../../examples/reference.json');
```
mbucknell: This location needs instance/reference.json as that is where we are putting it in the dockerfile process.

floodviz/thumbnail/thumbnail.js
```diff
@@ -24,14 +25,20 @@ if (args.length > 2) {
...
-	if (args[0] === '-css') {
-		style_path = args[1];
-	} else {
-		console.log('\nUnrecognized argument: ' + args[0] + '\n');
-		process.exit();
+	if (args[0]) {
+		if (args[0] === '-css') {
+			style_path = args[1];
+		} else {
+			console.log('\nUnrecognized argument: ' + args[0] + '\n');
```
mbucknell: Let's show the usage here too.

floodviz/thumbnail/thumbnail.js
```diff
@@ -44,19 +51,17 @@ jsdom.env(
 	[
 		'./floodviz/static/bower_components/d3/d3.js',
 		'./floodviz/static/bower_components/proj4/dist/proj4.js',
```
mbucknell: We don't need proj4 for the hydrograph, so this can be removed. Also remove the map div since we are only doing the hydrograph (that's above this on line 47).

floodviz/thumbnail/thumbnail.js
```diff
@@ -82,7 +87,7 @@ function convert(figure, window, css_path, filename) {
 		svg_string = inject_style(style_default, null, svg, window);
```
mbucknell: This comment is for line 78 where the additional css file is read in. Can this be done after line 30.? This gets the reading of the file done where the config parameter is read. You should be able to assign the value to style_ext and then you only need a single line to inject_style(style_default, style_ext, svg, window).

floodviz/thumbnail/thumbnail.js
ede0m: Gotcha!

floodviz/thumbnail/thumbnail.js
ede0m: For sure

floodviz/static/js/floodviz_onready.js
```diff
@@ -18,11 +18,12 @@ document.addEventListener('DOMContentLoaded', function (event) {
...
+		'display_ids': FV.hydrograph_display_ids,
 		'div_id': '#hydrograph'
 	};
 
 	var map_figure = FV.mapmodule(map_options);
-	var hydro_figure = FV.hydromodule(hydro_options);
+	var hydro_figure = hydromodule(hydro_options);
 
 	// Use frames to link interactions
 	var map_to_hydro = {
```
ede0m: I agree with this refactor. But I think it fits in better with a new ticket that would encapsulate the FV.default_display_id changes that still need to be made in map, for the sake of consistency across these figure modules. I would happily pick this new ticket up and make those changes tomorrow.

floodviz/thumbnail/thumbnail.js
ede0m: Good Idea! Thanks

floodviz/static/js/hydrograph.js
```diff
@@ -1,303 +1,310 @@
...
-		};
+	};
+	/**
+	 * Initialize the Hydrograph.
+	 *
+	 *@param {Object} linked_interactions - Object holding functions that link to another figure's interactions.
+	 *                                        Pass null if there are no such interactions to link.
+	 *        @prop 'hover_in' - linked interaction function for hover_in events on this figure.
+	 *        @prop 'hover_out' - linked interaction function for hover_out events on this figure.
+	 *        @prop 'click' - linked interaction function for click events on this figure.
```
mbucknell: Add comment for @param data

PR #94, opened 2017-08-01
Switch rdb parser fv 41

jameeters:
Converts map and peakflow utils to use the new RDB parser and changes their unittests to match.


floodviz/map_utils.py
```diff
@@ -1,69 +1,53 @@
...
-        return None
+    url = url_prefix + 'site/'
 
-    if req.text == "":
-        print("Service call returned no data")
-        return []
+    params = {
+        'format': 'rdb',
+        'sites': id_input_string,
+        'siteStatus': 'all',
```
ede0m: could change variable naming style here if we care. site_status

floodviz/peak_flow_utils.py
```diff
@@ -1,104 +1,125 @@
...
+    }
+    content = utils.parse_rdb(url_prefix, params)
+    if content is None:
+        print('rdb parser has returned no data to peakflow util req_peak_dv_data')
+        peaks = None
 
+    else:
+        # Find the name of that column whose name changes
+        for k in content[0].keys():
+            colname = re.match(r'\d+_00060_00003$', k)
```
ede0m: nice re. I'm just hoping that this column header doesn't vary its name for some reason.

floodviz/peak_flow_utils.py
```diff
@@ -1,104 +1,125 @@
...
+            # if a match has been found
+            if colname is not None:
+                colname = colname.string
+                break
+        # (no break) means that this column has not been found
+        else:
```
ede0m: Is this just a last ditch effort to get the value based on position heuristic if re doesn't work?

floodviz/peak_flow_utils.py
```diff
@@ -1,104 +1,125 @@
...
-                continue
-            line = line.split('\t')
-            year = line[2].split('-')[0]
-            # remove duplicate years
-            if year in seen:
-                continue
-            else:
+        for point in peak_data:
+            # I am not worried about years that are not four digits long
+            year = re.match(r'^(\d{4}).*', point['peak_dt']).group(1)
```
ede0m: Not sure if we need this or not, it doesn't hurt to have this check I suppose but I didn't encounter any years that were not 4 digits.

floodviz/peak_flow_utils.py
jameeters: This just deals with the first few digits changing, so any more variation than that will cause problems.

floodviz/peak_flow_utils.py
jameeters: Yep, that's exactly what this is.

floodviz/peak_flow_utils.py
jameeters: Yeah, this may be an overuse of regular expressions, so I'd be open to removing it.

mbucknell: Does seem like overkill, but fine to leave in.

floodviz/tests/test_peak_flow_utils.py
```diff
@@ -2,235 +2,245 @@
...
+                             'USGS	05481950	2008-06-01		7800		14.51'
+
+        self.correct_outupt = self.correct_historic_peaks_output
+
+        self.prefix = 'http://example.com/'
+        self.end_date = '2008-07-05'
+        self.site = '05481950'
+        self.mock_url = self.prefix + '?site_no=' + self.site + '&agency_cd=USGS&format=rdb&end_date=' + self.end_date
+
+    def test_bad_status_code(self):
```
mbucknell: To create a true unit test, you would not be mocking something that is called by a function that is called by the unit under test. Instead you would mock utils.parse_rdb. You can leave this as is. But something to consider for the future.

floodviz/tests/test_peak_flow_utils.py
```diff
@@ -2,235 +2,245 @@
...
+                                     {'peak_dt': '2003-05-06', 'peak_va': '3160'},
+                                     {'peak_dt': '2004-05-24', 'peak_va': '7300'},
+                                     {'peak_dt': '2005-05-14', 'peak_va': '2530'},
+                                     {'peak_dt': '2006-09-14', 'peak_va': '1740'},
+                                     {'peak_dt': '2007-02-02', 'peak_va': '850'},
+                                     {'peak_dt': '2007-04-27', 'peak_va': '6850'},
+                                     {'peak_dt': '2008-06-01', 'peak_va': '7800'}, ]
+
+
+class TestReqPeakDataPeakflow(PeakflowBaseTest):
```
mbucknell: Please change the name of the class to TestReqPeakData. That is the best practice we have established here.

floodviz/tests/test_peak_flow_utils.py
```diff
@@ -2,235 +2,245 @@
...
+            m.get(self.mock_url, status_code=404)
+            self.assertEqual(req_peak_data(self.site, self.end_date, self.prefix), None)
+
+    def test_valid_data(self):
+        with requests_mock.Mocker() as m:
+            m.get(self.mock_url, text=self.nwis_response)
+            self.assertEqual(req_peak_data(self.site, self.end_date, self.prefix), self.correct_outupt)
+
+
+class TestReqPeakDVDataPeakflow(PeakflowBaseTest):
```
mbucknell: Please change the name to TestReqPeakDVData

floodviz/tests/test_peak_flow_utils.py
```diff
@@ -2,235 +2,245 @@
...
+            m.get(self.mock_url, status_code=404)
+            self.assertEqual(req_peak_dv_data(self.site, self.date, self.prefix), None)
+
+    def test_valid_data(self):
+        with requests_mock.Mocker() as m:
+            m.get(self.mock_url, text=self.nwis_response)
+            self.assertEqual(req_peak_dv_data(self.site, self.date, self.prefix), self.correct_output)
+
+
+class TestParsePeakDataPeakflow(PeakflowBaseTest):
```
mbucknell: Please change the name to TestParsePeakData

PR #96, opened 2017-08-02
Json ld fv 32

jameeters:
This is just a first attempt at generating JSON-LD to describe the date range and location of the flood and the gage metadata. I think the generator itself is good, but I haven't found a good way to structure the output.


floodviz/linked_data_utils.py
```diff
@@ -0,0 +1,150 @@
...
+        lat = [bbox[1], bbox[3]]
+        minlat = min(lat)
+        maxlat = max(lat)
+        minlon = min(lon)
+        maxlon = max(lon)
+
+        self.location = [minlat, minlon, maxlat, maxlon]
+
+    def assemble(self):
+        """
```
ede0m: maybe have a check here to see if things like location and gauges have been set yet? Just thinking in the context of them reusing this class in a separate project. Or you could do this in the private methods you call here.

floodviz/linked_data_utils.py
```diff
@@ -0,0 +1,150 @@
...
+        Put together all data
+        :return: return a JSON-LD-like dictionary
+        """
+        self.ld['about'] = self._assemble_event()
+        gages = self._assemble_all_gages()
+        for g in gages:
+            self.ld.update(g)
+        print(gages)
```
ede0m: Take out print

floodviz/views.py
```diff
@@ -7,26 +7,37 @@
 from . import map_utils
 from . import peak_flow_utils
 from . import REFERENCE_DATA as ref
+from .linked_data_utils import LinkedData
 
 url_nwis_prefix = app.config['NWIS_SITE_SERVICE_ENDPOINT']
+linked_data = LinkedData()
```
ede0m: Setters vs passing things like bbox and dates to the actual object. I'm not sure which is preferred here, but just noting the other option.

floodviz/views.py
```diff
@@ -57,8 +67,10 @@ def _peakflow_helper():
     peak_data = peak_flow_utils.parse_peak_data(historic_peaks, daily_value_data)
     return peak_data
 
+
 def _map_helper():
     site_data = map_utils.site_dict(ref['site_ids'], app.config['NWIS_SITE_SERVICE_ENDPOINT'])
+    linked_data.set_gages(site_data)
```
ede0m: After seeing this, I think you made the right decision regarding my previous comment about setters vs passing as object params.

floodviz/linked_data_utils.py
```diff
@@ -0,0 +1,150 @@
...
+                "alternateName": "USGS"
+            },
+        })
+
+        self.gages = []
+        self.dates = {}
+        self.location = []
+
+    @staticmethod
+    def _blank_thing(typename):
```
ede0m: Maybe change to something like 'new_ld_frame' or something along those lines.

jameeters: The base of all the schema.org ontologies is actually called 'thing', and I think this is easier to remember.

floodviz/linked_data_utils.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+class LinkedData:
+    """
+    Generates JSON-LD based on gages and the flood event.
+    """
+    def __init__(self):
+        self.ld = self._blank_thing("WebSite")
+        self.ld.update({
+            "name": "Active flood visualization",
```
mbucknell: I think we should be getting the name and about through config variables. This might not be ideal in the end workflow but let's start with that.

floodviz/linked_data_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        """
+        Wrap the data on the event into a dictionary
+        :return: JSON-LD-like dict representing the event
+        """
+        event = self._blank_thing("Event")
+        if self.location and self.dates:
+            event.update({
+                "@context": "http://schema.org",
+                "@type": "Event",
+                "name": "FLOOD EVENT NAME",
```
mbucknell: I would use app.config['TITLE'] here for this.

floodviz/linked_data_utils.py
jameeters: Actually I'm replacing the 'about' with the flood event later, so I'll just remove that setting.

PR #98, opened 2017-08-02
Add ability to generate thumbnail during Docker build

skaymen:


Dockerfile
```diff
@@ -25,7 +26,8 @@ RUN rm -f /etc/apt/sources.list.d/chris-lea-node_js-jessie.list \
 && curl https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - \
 && echo 'deb https://deb.nodesource.com/node_6.x jessie main' > /etc/apt/sources.list.d/nodesource.list \
 && apt-get update \
-&& apt-get install -y nodejs
+&& apt-get install -y nodejs \
+&& apt-get install -y libfontconfig
```
jameeters: This isn't a part of the node installation, so I think it should be its own `RUN` command, included in a more related command, or added to the list at  the top of the file.

Dockerfile-DOI
```diff
@@ -27,7 +28,8 @@ RUN rm -f /etc/apt/sources.list.d/chris-lea-node_js-jessie.list \
 && curl --cacert DOIRootCA2.cer https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - \
 && echo 'deb https://deb.nodesource.com/node_6.x jessie main' > /etc/apt/sources.list.d/nodesource.list \
 && apt-get update \
-&& apt-get install -y nodejs
+&& apt-get install -y nodejs \
+&& apt-get install -y libfontconfig
```
jameeters: Same as in the other dockerfile: I think this should be at the top of the file, in its own command, or included with a more related command

README.md
```diff
@@ -17,7 +17,7 @@ the frozen files.
 **To build the image:**
 1. Navigate to your project directory.
 1. Copy `DOIRootCA2.cer` into this directory.
-1. Run `docker build --build-arg config=examples/iowa.py --build-arg ref=examples/reference.json -t flood:latest -f Dockerfile-DOI .`. This will create the docker image and name it `flood`.
+1. Run `docker build --build-arg config=examples/iowa.py --build-arg ref=examples/reference.json --build-arg thumbnail=true  -t flood:latest -f Dockerfile-DOI .`. This will create the docker image and name it `flood`.
```
mbucknell: Seems wrong to set thumbnail twice essentially, one for generating the data and one for generating the actual thumbnail. I think we could do something similar to what I did to set the FREEZER_BASE_URL (see config.py in project home) using an environment variable. So pass in thumbnail as a build arg, set an environment variable in the docker container to control he value of THUMBNAIL in config.py before running the freeze command. What do you think?

examples/iowa.py
```diff
@@ -24,4 +24,4 @@
 }
 
 # Thumbnail Support
-THUMBNAIL = False
\ No newline at end of file
+THUMBNAIL = True
```
mbucknell: This can be removed if we implement the change I suggested above.

PR #100, opened 2017-08-02
Google analytics fv 57

ede0m:


floodviz/static/js/analytics_helper.js
```diff
@@ -0,0 +1,45 @@
...
+	Array.prototype.forEach.call(links, function (element) {
+		element.addEventListener('click', function () {
+			ga_send_event('Links', 'click_link', element.href);
+		});
+	});
+
+});
+
+
+function ga_send_event(category, action, label) {
```
skaymen: Could use function comments.

floodviz/static/js/hydrograph.js
```diff
@@ -205,6 +208,10 @@ var hydromodule = function (options) {
 				self.linked_interactions.hover_in(d.data.name, d.data.key);
 				self.activate_line(d.data.key);
 				self.series_tooltip_show(d);
+				if (!hydro_moused_over) {
```
skaymen: Couldn't hurt to throw comments in for this process in hydrograph, map, etc

floodviz/static/js/analytics_helper.js
```diff
@@ -0,0 +1,45 @@
...
+	Array.prototype.forEach.call(links, function (element) {
+		element.addEventListener('click', function () {
+			ga_send_event('Links', 'click_link', element.href);
+		});
+	});
+
+});
+
+// Wrapper function for Google Analytics explicit send
+function ga_send_event(category, action, label) {
```
mbucknell: Could we assign this to the FV namespace? Also I think you can add this directly to index.html. Probably below the code that sets up the FV namespace.

floodviz/static/js/map.js
```diff
@@ -56,6 +56,8 @@
 			var maptip = d3.select('body')
 				.append('div')
 				.attr('id', 'maptip');
+			// Google Analytics Boolean Trackers
+			var map_moused_over_gauge = {};
```
mbucknell: Nit: Please use 'gage' rather than "gauge". It's the standard spelling for USGS. Let's not rock the boat!

floodviz/static/js/map.js
```diff
@@ -230,6 +232,7 @@
 						}
 					});
 					self.linked_interactions.click(selected);
+					ga_send_event('Map', 'drag_select', selected);
```
mbucknell: The variable selected is an array and the function expects a string. You can easily convert to a string by doing selected.join(',')

floodviz/static/js/map.js
```diff
@@ -301,13 +304,19 @@
...
+							map_moused_over_gauge[d.properties.id] = true;
+						}
 					})
 					.on('mouseout', function (d) {
 						self.site_tooltip_remove();
 						self.linked_interactions.hover_out(d.properties.id);
 					})
 					.on('click', function (d) {
 						toggle_hydrograph_display(d.properties.id);
+						ga_send_event('Map', 'gauge_click_on', d.properties.id);
```
mbucknell: nit: spelling again.

floodviz/static/js/analytics_helper.js
```diff
@@ -0,0 +1,45 @@
...
+ga('create', 'UA-103831764-1', 'auto');
+ga('send', 'pageview');
+
+
+// GA for DOM content not related to figures.
+document.addEventListener('DOMContentLoaded', function (event) {
+
+	// Social Media Click events for GA
+	var sm_icons = document.getElementsByClassName('icon');
+	Array.prototype.forEach.call(sm_icons, function (element) {
```
mbucknell: Is there a reason you are using Array.prototype.forEach.call rather than sm_icons.forEach? This is also one place where I think it's clearer to add a 'onclick' property to each icon in index.html: so something like 
<img src="static/images/faceblook.pn" onclick="ga_send_event('Social_Media', 'click_icon', 'Facebook')". Then this code can be eliminated

ede0m: sm_icons and links are HTMLCollections which do not have the forEach function available to them.

 I was trying to keep the GA code as separated as possible by putting this logic in this file.

floodviz/static/js/analytics_helper.js
ede0m: I was trying to separate GA code into this file instead of keeping it all in the index.html. Maybe we can talk about this at standup?

floodviz/static/js/map.js
ede0m: Good Catch. This definitely threw off the label in the GA views when I was checking.

floodviz/static/js/analytics_helper.js
mbucknell: One reason we put it directly in the html is it's easy to see by inspecting the page source whether google analytics has been implemented. I'd like to stick that. It is an exception to how do things otherwise.

ede0m: Since I am moving the other parts of the analytics_helper file, should I move the google's tracking code to the html as well?

floodviz/templates/index.html
```diff
@@ -11,9 +11,39 @@
...
+			a.src = g;
+			m.parentNode.insertBefore(a, m)
+		})(window, document, 'script', 'https://www.google-analytics.com/analytics.js', 'ga');
+
+		ga('create', 'UA-103831764-1', 'auto');
+		ga('send', 'pageview');
+
+		// Global Namespace for project - FV
+		var FV = FV || {};
+
```
ede0m: This needed to be defined up here so we have the ability to call FV.ga_send_event on html content (links, icons, etc)

PR #101, opened 2017-08-03
Keep hydrograph tooltip visible fv 55

jameeters:
This is just part of FV-55: it keeps the hydrograph tooltip from being cut off by the left of right edges of the svg.
I made a jira sub-task for this: FV-77

### DIFF
--- a/floodviz/static/js/hydrograph.js
+++ b/floodviz/static/js/hydrograph.js
@@ -121,6 +121,7 @@ var hydromodule = function (options) {
 			.append('g')
 			.attr('transform', 'translate(' + margin.left + ',' + margin.top + ')');
 
+		// Save the locations of the edges of the visible svg
 		state.edges = {
 			'l': - (margin.left + margin.right),
 			'r': width + margin.right,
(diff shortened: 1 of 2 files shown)


floodviz/static/js/hydrograph.js
```diff
@@ -14,44 +13,45 @@
...
 
+	var state = {};
+
 	var default_display_ids = null;
 	var timer = null;
 	var dblclick_armed = false;
 
-	var margin = {top: 60, right: 0, bottom: 30, left: 35};
-	var height = 500 * (options.height / options.width) - margin.top - margin.bottom;
+	var margin = {top: 60, right: 0, bottom: 30, left: 50};
```
skaymen: 40 might work better for a left margin so it doesn't appear off-center with the text box

floodviz/static/js/hydrograph.js
```diff
@@ -212,6 +217,7 @@ var hydromodule = function (options) {
 				self.series_tooltip_remove(d.data.key);
 			})
 			.on('click', function (d) {
+				console.log(d3.mouse(this));
```
mbucknell: I'm assuming that this will be removed before the complete ticket is done,

PR #102, opened 2017-08-03
Google Analytics - parameterize ga id 

ede0m:
Just wanted to add this change quick.

**EDIT:**
This PR now contains a parameterized version of handling the GA_ID along with the previous change I made to handle link event labels.

### DIFF
--- a/config.py
+++ b/config.py
@@ -58,7 +58,7 @@ MAP_CONFIG = {
 THUMBNAIL = False
 
 # Google Analytics
-GA_ID = None
+GA_ID = "null"
 
 deployed_url_base = os.environ.get('DEPLOYED_BASE_URL')
 if deployed_url_base:
--- a/config.py
+++ b/config.py
@@ -54,9 +54,6 @@ MAP_CONFIG = {
     'scale': None,
 }
 
-# Thumbnail Support
-THUMBNAIL = False
-
 # Google Analytics
 GA_ID = ""
 
--- a/config.py
+++ b/config.py
@@ -57,6 +57,9 @@ MAP_CONFIG = {
 # Thumbnail Support
 THUMBNAIL = False
 
+# Google Analytics
+GA_ID = None
+
 deployed_url_base = os.environ.get('DEPLOYED_BASE_URL')
 if deployed_url_base:
     FREEZER_BASE_URL = deployed_url_base
--- a/config.py
+++ b/config.py
@@ -58,7 +58,7 @@ MAP_CONFIG = {
 THUMBNAIL = False
 
 # Google Analytics
-GA_ID = "null"
+GA_ID = ""
 
 deployed_url_base = os.environ.get('DEPLOYED_BASE_URL')
 if deployed_url_base:
--- a/examples/iowa.py
+++ b/examples/iowa.py
@@ -23,9 +23,6 @@ PEAK_META = {
     'width': '7.5'
 }
 
-# Thumbnail Support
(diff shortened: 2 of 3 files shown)


floodviz/templates/index.html
```diff
@@ -16,6 +16,16 @@
 
 	</script>
 	<script type="text/javascript">
+
+		// Google Analytics ID
```
mbucknell: There is an easier way to do this. Use a template if block around the code that sets up the google analytics. See http://jinja.pocoo.org/docs/2.9/templates/#tests. This will set up google analytics except without any code which is not good. You will also want to use an if block in ga_send_event to do nothing if GA_ID is null.

examples/iowa.py
```diff
@@ -24,4 +24,7 @@
 }
 
 # Thumbnail Support
-THUMBNAIL = False
\ No newline at end of file
+THUMBNAIL = False
+
+# Google Analytics
+GA_ID = '103831764'
```
mbucknell: Typically we put in the whole id so this would be UA-103831764-1

ede0m: calling {{ config.GA_ID }} on the whole id only returns UA. Any idea why this is happening?

floodviz/templates/index.html
```diff
@@ -10,40 +11,45 @@
...
-		{{ linked_data | tojson | safe }}
-
-	</script>
+	<script type="application/ld+json"> {{ linked_data | tojson | safe }} </script>
+	<script> var FV = FV || {}; </script>
+	{% if config.GA_ID != "null" %}
+		<script> FV.GA_RUN = true; </script>
+	{% else %}
+		<script> FV.GA_RUN = false; </script>
+	{% endif %}
```
ede0m: Not a huge fan of how I had to implement this. But wrapping the jinja condition around the entire GA script would cause event errors later with figure interactions. I thought about moving the ga_send_event, but it is called directly from html which makes it required up top here. This at least avoids importing the ga_analytics script when it is not needed.

config.py
```diff
@@ -58,7 +58,7 @@
 THUMBNAIL = False
 
 # Google Analytics
-GA_ID = None
+GA_ID = "null"
```
mbucknell: Why not just pass in the empty string ('') ?

floodviz/templates/index.html
mbucknell: Let's work on this together once you get in. You shouldn't have to create the variable FV.GA_RUN. Rather use the if block to conditionally include the google analytics script.

ede0m: How are we supposed to handle calls to ga_send_event when the id is null?

config.py
```diff
@@ -54,6 +54,12 @@
     'scale': None,
 }
 
+# Thumbnail Support
+THUMBNAIL = False
```
mbucknell: This has not been merged correctly. This needs to be fixed.

examples/iowa.py
```diff
@@ -22,3 +22,10 @@
     'height': '4',
     'width': '7.5'
 }
+
+# Thumbnail Support
+THUMBNAIL = False
```
mbucknell: This line should not be in here. It looks like there was a merge issue with this as well.

config.py
ede0m: Why are we not using this now? I wasn't aware of this.

ede0m: Just talked to stan, he filled me in. I will remove this.

PR #103, opened 2017-08-03
USGS Watermark and css

ede0m:
I put the watermark in the upper right. I am assuming there will be some discussion on where we actually want to put it. 
hydrograph


floodviz/static/js/hydrograph.js
```diff
@@ -122,6 +122,46 @@ var hydromodule = function (options) {
 			.attr('transform',
 				'translate(' + margin.left + ',' + margin.top + ')');
 
+		// Watermark
+		var watermark = svg.append('g')
+			.attr('id', 'usgs-watermark')
+			.attr('transform', 'translate(' + 2 / 3 * width + ',0' + ')scale(0.20)');
+		watermark.append('path')
+			.attr('d', 'm234.95 15.44v85.037c0 17.938-10.132 36.871-40.691 36.871-27.569 0-40.859-14.281-40.859-' +
```
skaymen: might be better to store this in a var or pass in from elsewhere

PR #107, opened 2017-08-07
Missing data fv 67

ede0m:
This is not complete. I am still waiting on some information to be able to handle gauges that sample at a frequency under 15 minutes. 

**Edit:**

We now only consider data to be missing if the gap between adjacent datapoints is greater than 30 minutes. All dummy data points will be added in 15 minute increments. This does not affect the figure visual.


floodviz/hydrograph_utils.py
```diff
@@ -37,8 +41,21 @@ def parse_hydrodata(jdata):
                 dt = datetime.strptime(date + ' ' + t, '%Y-%m-%d %H:%M:%S')
                 # Convert to milliseconds for use with d3 x axis format
                 dt_ms = time.mktime(dt.timetuple()) * 1000
+                # handle missing data
+                if prev_date_ms:
```
skaymen: It might help to have more detailed comments to explain variables and steps of the calculation.

floodviz/static/js/hydrograph.js
```diff
@@ -213,7 +240,8 @@ var hydromodule = function (options) {
 		voronoi_group = svg.append('g')
 			.attr('class', 'voronoi');
 		voronoi_group.selectAll('path')
-			.data(voronoi.polygons(d3.merge(dataNest.map(function (d) {
+			.data(voronoi.polygons(d3.merge(dataNest_voronoi.map(function (d) {
+				//console.log(d.values);
```
skaymen: This should probably be taken out before merging

floodviz/hydrograph_utils.py
```diff
@@ -19,13 +20,17 @@ def parse_hydrodata(jdata):
 
     """
     all_series_data = []
-    
+    # gages can collect and send data every 15 minutes
```
mbucknell: This comment is not true. Maybe instead indicate that this code will fill in data gaps of more the 30 minutes

ede0m:
Not sure why tests are failing here. nosetests are passing locally.

floodviz/tests/test_hydrograph_utils.py
```diff
@@ -2,41 +2,79 @@
...
-                                                                                                                                                                                                                                                                                                                                                                                                                                      [{"name": "Statistic", "optionCode": "00000"}]}, "note": [], "noDataValue": -999999.0, "variableProperty": [], "oid": "45807197"}, "values":
-                                                           [{"value": [{"value": "446", "qualifiers": ["A"], "dateTime": "2008-05-20T00:00:00.000-05:00"}]}]}}}
-
-        ## Returned value is different than the mocked request response because req_hydrodata accessess two extra levels of the dict before returning.
+        self.mock_response = {
+            "value": {"timeSeries": {"sourceInfo": {"siteName": "Black Hawk Creek at Hudson, IA", "siteCode":
+                [{"value": "05463500", "network": "NWIS", "agencyCode": "USGS"}], "timeZoneInfo": {"defaultTimeZone":
+                {
```
mbucknell: The indenting makes this hard to read. Could the indentation be cleaned up? Look at test_map_utils.py for an example.

floodviz/tests/test_hydrograph_utils.py
```diff
@@ -59,21 +97,43 @@ def test_bad_url(self):
...
-            "timeZoneInfo": {"defaultTimeZone": {"zoneOffset": "-06:00","zoneAbbreviation": "CST"},"daylightSavingsTimeZone": {
-                "zoneOffset": "-05:00","zoneAbbreviation": "CDT"},"siteUsesDaylightSavingsTime": True},"geoLocation": {
-                "geogLocation": {"srs": "EPSG:4326","latitude": 42.4077639,"longitude": -92.4632451},"localSiteXY": []},
-            "note": [],"siteType": [],},"variable": {"variableCode": [],"variableName": "Streamflow, ft&#179;/s",
-                                                     "variableDescription": "Discharge, cubic feet per second","valueType": "Derived Value","unit": {"unitCode": "ft3/s"},
-                                                     "note": [],},"values": [{"value": [{
-            "value": "446","qualifiers": ["A"],"dateTime": "2008-05-20T00:00:00.000-05:00"},{
-            "value": "446","qualifiers": ["A"],"dateTime": "2008-05-20T00:15:00.000-05:00"},{
-            "value": "446","qualifiers": ["A"],"dateTime": "2008-05-20T00:30:00.000-05:00"}]}]}]
+            "siteCode": [{"value": "05463500", "network": "NWIS", "agencyCode": "USGS"}],
```
mbucknell: Let's cleanup the indentation here too.

floodviz/hydrograph_utils.py
```diff
@@ -37,8 +41,27 @@ def parse_hydrodata(jdata):
...
+                      We only consider data to be missing if the gap between adjacent data points 
+                      is greater than 30 minutes. All dummy data points will be added in 15 minute increments.
+                    """
+                    if gap > gap_threshold:
+                        num_dummy_points = gap/increment_ms - 1
+                        added = 0
+                        while added < num_dummy_points:
+                            # correct time for this dummy point
+                            new_dt_ms = prev_date_ms + (increment_ms * (added + 1))
+                            all_series_data.append({'key': site_id, 'name': 'no_data', 'date': date, "time": t,
```
mbucknell: Is there a reason you are assigning 'no_data' to 'name'? I would expect this to contain site_name?

floodviz/tests/test_hydrograph_utils.py
ede0m: This is what pycharm > code > reformat code gave me :\

floodviz/hydrograph_utils.py
ede0m: I wanted an indication that there was no data available for site number for these points.

mbucknell: Is this used in the javascript?

floodviz/tests/test_hydrograph_utils.py
mbucknell: Yeah the default formatting is not good. It helps a little if you set the line length to 80 I think. You may have to do this one more by hand. Generally, it's easier to read with one property per line. Don't worry about it now. But if you have time this week, it would be nice to pretty this up.

floodviz/hydrograph_utils.py
ede0m: No it is used in python testing however.

floodviz/tests/test_hydrograph_utils.py
ede0m: Sure thing! I'll get it done before the end of the week.

floodviz/hydrograph_utils.py
mbucknell: You need to find another way to do the test then. The siteName should be the siteName whether there is data or not.

floodviz/tests/test_hydrograph_utils.py
```diff
@@ -194,6 +169,12 @@ def test_valid_data_missing_count(self):
     def test_valid_data_missing_value(self):
         ret = parse_hydrodata(self.mock_data_missing_points)
         for dp in ret:
-            if dp.get('name') is 'no_data':
-                self.assertEqual(dp.get('value'), 'NA')
+            try:
```
mbucknell: You should know exactly where the dummy data will be in the list. You should look for it there. This test would succeed if the first element in the list was null for instance and doesn't check to see if the values where filled in every 15 minutes for the gap.

ede0m: Yes, because of mock_data_missing_points I do know where the NA value is. I was really trying to not have to do something like ret[3] here if that's what you were thinking? It just seems like a bad idea to conform the tests to a hard coded position, when in theory that point could be at almost any position within the set.

mbucknell: The point is you should know where the NA's are. These are tests and you should know exactly what you are expecting.  So yes. We've hardcode the intervals in the function itself so it's reasonable to have to change the tests if those constants are updated.

PR #109, opened 2017-08-08
changed request url builder key dict

ede0m:


ede0m:
Not sure why the travis check failed here. Anyone have an idea?

PR #110, opened 2017-08-08
Combine tooltips fv 83

jameeters:
Combines the hydrograph and map tooltip logic into a single new module.

### DIFF
--- a/floodviz/static/js/floodviz_onready.js
+++ b/floodviz/static/js/floodviz_onready.js
@@ -7,6 +7,8 @@ document.addEventListener('DOMContentLoaded', function (event) {
 	};
 
 	const tooltip = tooltip_module(tooltip_options);
+	FV.show_tooltip = tooltip.show_tooltip;
+
 	// Map options
 	var map_options = {
 		'height': FV.mapinfo.height,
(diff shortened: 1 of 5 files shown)


floodviz/static/js/floodviz_onready.js
```diff
@@ -1,6 +1,7 @@
 document.addEventListener('DOMContentLoaded', function (event) {
 	'use strict';
 
+	const tooltip = tooltip_module(null);
```
mbucknell: If it has no options, you can just call like this ```tooltip_module();```

floodviz/static/js/tooltip.js
```diff
@@ -0,0 +1,128 @@
+'use strict';
+const tooltip_module = function (options) {
```
mbucknell: This isn't really using the module pattern. This is really just a single stateless function show_tooltip. I would be okay just exposing the show_tooltip function. Maybe something like FV.show_tooltip

jameeters:
I tested the thumbnail generation and it still works during docker build, so I don't think the thumbnail generation ever calls the mouseover handler.

PR #115, opened 2017-08-09
Fix map display ids fv 66 and fv 85

jameeters:


floodviz/static/js/floodviz_onready.js
```diff
@@ -21,7 +21,8 @@ document.addEventListener('DOMContentLoaded', function (event) {
 		'rivers_data': FV.mapinfo.rivers_data,
 		'ref_data': FV.mapinfo.ref_data,
 		'site_data': FV.mapinfo.site_data,
-		'div_id': '#map'
+		'div_id': '#map',
+		'hydrograph_display_ids': FV.hydrograph_display_ids
```
mbucknell: I would prefer this be called site_display_ids just to emphasize that they don't have to be associated with a hydrograph.

floodviz/static/js/hydrograph.js
```diff
@@ -351,6 +351,9 @@ var hydromodule = function (options) {
 	 */
 	self.change_lines = function (new_display_ids) {
 		options.display_ids = new_display_ids;
+		if (!disableInteractions){
+			self.linked_interactions.update_map_display_ids(new_display_ids);
```
mbucknell: Add comments for this function and accent_on_map to the doc comments for init.

PR #121, opened 2017-08-10
Reset button appears after changing hydrograph

skaymen:


floodviz/static/css/hydrograph.css
```diff
@@ -102,3 +102,10 @@
     stroke-linejoin: round;
     stroke-miterlimit: 10.00;
 }
+
+.btn {
```
jameeters: It might be better to style it by ID rather than class.

floodviz/templates/index.html
```diff
@@ -59,7 +59,7 @@
 				</div>
 			</div>
 			<div class="col-sm-6 col-centered">
-				<div id="hydrograph"></div>
+				<div id="hydrograph"><input class="btn btn-primary pull-right" id="button" type="reset" value="Reset"></div>
```
jameeters: 'reset-button' would probably be a better id.

floodviz/static/js/hydrograph.js
```diff
@@ -228,6 +230,16 @@ var hydromodule = function (options) {
 			.style("font-size", "14px")
 			.text("Discharge (cubic feet per second)");
 
+		var button = d3.select(".btn");
+		button.on('click', function() {reset_hydrograph();});
+
+		if (default_display_ids.toString() === options.display_ids.toString()) {
```
ede0m: why do we have to compare by string here?

floodviz/templates/index.html
skaymen: Fixed.

floodviz/static/css/hydrograph.css
skaymen: Fixed.

floodviz/static/js/hydrograph.js
mbucknell: This comparision makes me nervous. You are relying on the ids being in the same order. Okay for now, but it seems to me we are doing far too much in this update procedure. It is over 150 lines. If we had more time, I would look at refactoring the hydrograph module.

PR #126, opened 2017-08-11
Icons are now links. Added reddit icon

skaymen:


floodviz/templates/index.html
```diff
@@ -36,12 +36,21 @@
 	<div class="main-content container">
 		<div class="col-xs-offset-11">
 			<div class="icon-div">
-				<img src="static/images/facebook.png" class="icon"
+				<a href="https://www.facebook.com/sharer/sharer.php?u=" + {{ config.PAGE_URL | safe }}>
```
mbucknell: Is this supposed to be the url of this page? If so you should be using url_for. Otherwise this won't work when deployed to CHS

